PR #31, opened 2017-01-26
Pk migrate type

pkelly:


pkelly:
@GeorgeTaveras1231: I added the Bootstrap custom var overrides and am working on updating core/typography.

I want to move over the woff fonts to get rid of the inlined base64 ones. You have any thoughts on how we should include them with webpack?

GeorgeTaveras1231:
@pkelly I think I'm still unsure about what the best strategy to deliver fonts is, I like using the data urls because it makes the css self-contained. This makes it easier for us (The developers of union) because we just publish the css module as npm packages without any other concerns. On the other end, we could distribute the fonts via a CDN and write the css in a way that it makes references to the CDN. This approach only concerns me because it makes the deployment/publishing process a little more complex. This means we have to set up the hosting infrastructure and the build steps to deploy the fonts.

pkelly:
I don't want to continue to inline the fonts because they are quite large and since CSS rendering blocks page render it currently causes a serious delay. Additionally it prevents us from using woff2 (better compression) with woff as a fallback.

Hosting the fonts from the same CDN url across apps does give us a caching benefit. Since fonts very rarely change maybe this is the simplest strategy for the time being.

GeorgeTaveras1231:
@pkelly thanks for clarifying some of the benefits I think I'm sold on doing that. 

As far as how we should include them with webpack, it seems webpack will play a small role in distributing the fonts; we just need to reference the font's url in the css.

pkelly:
Found a place for the fonts to live which is backed by the cdn.
Example: https://static.xoedge.com/union/fonts/tisa-light.woff2

packages/core/typography/src/index.scss
```diff
@@ -1,22 +1,149 @@
 @import './fonts';
 @import './bootstrap-build';
```
pkelly: @GeorgeTaveras1231: Is there a way I can reference `_global.scss` in core/bootstrap?

Since I added the Bootstrap config vars in that package I need to reference them in typography too.

GeorgeTaveras1231: There isn't a way currently, this is definitely something we should figure out. 

You could consider bringing the `_global.scss` file closer to the top level of the project and change the sass build configs on each of the packages to look in the centralized place.

GeorgeTaveras1231:
> Found a place for the fonts to live which is backed by the cdn.

@pkelly is this in our aws s3? if so, which bucket? I'm thinking whether we should create a new bucket for union stuff.

packages/core/typography/src/index.scss
pkelly: Once these packages are published on NPM, I could add a dependency on `@union/bootstrap` and pull in the sass that way right?

For now I'll just move the _global.scss file higher up.

pkelly:
I finished up the type changes. Circle-ci fails, but it appears that is cause the https://github.com/xogroup/union/pull/35 needs to be merged first.

@GeorgeTaveras1231: Could you review this when you get a chance? If things look good I can merge it.

pkelly:
I'm able to replicate the circleci error now. I had to run `npm run clean`

PR #35, opened 2017-02-01
Test Setup

### LINKED ISSUE #36: Setup CI services
##### Things to do:

- [x] *Decide on a CI service. Some options:*
- Circle CI
- ~~Travis~~

> I chose circle because:
> - It took less than 10 minutes to set up
> - Is used by other xo projects

- [x] *Create pull request build*

GeorgeTaveras1231:
Closes #36


package.json
```diff
@@ -23,7 +24,8 @@
     "start": "pm2 start $npm_package_config_process_config_file",
     "status": "pm2 status",
     "stop": "pm2 stop all",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "tdd": "karma start",
```
pkelly: Maybe `test:watch` would be a better indicator of what this command does

pkelly: Might be nice to have `test:chrome` too for `karma start --browsers=Chrome` if you need to use the debugger for your specs

GeorgeTaveras1231: Totally agree, thanks for the feedback!

PR #40, opened 2017-02-06
Add simple 404 page

GeorgeTaveras1231:
This doesn't look pretty at all but makes it easier to tell when you visit the wrong URL.


pattern-library/docs/src/components/Errors/404.jsx
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    if (route.path !== '*') {
+      routes.push(<li key={path}><a href={path}>{path}</a></li>);
+    }
+  });
+
+  return (
+    <div>
+      {"This route doesn't exist."}
```
pkelly: Could this error message be more human readable? Maybe `Page not found`

I like how the routes are listed out. Neat way to expose the sitemap 😀

Could we wrap the error message in an `<h1>` to make it more visible.

pattern-library/docs/src/components/Errors/404.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+      <ul>
```
pkelly: Could we add `type.listUnstyled` to this to get rid of the bullets?

PR #55, opened 2017-02-14
Pk add type doc page

GeorgeTaveras1231:


GeorgeTaveras1231:
@pkelly just a quick explanation of the major changes in the `typography` package.

There are a couple of files that are included in the package and this is what they provide:

```js
// This includes the rules that have element selectors (and considered global). 
// Nothing meaningful is exported. You can see an example of this here:
// https://github.com/xogroup/union/blob/master/pattern-library/docs/src/index.jsx#L7
import '@xo-union/typography/globals'; 

// This includes the css modules so the class map is exported
import typeModule from '@xo-union/typography';

// These other ones would be less relevant to consumers but are available
// and mostly used by the two modules above

// This includes the font names and urls
import typeData from '@xo-union/typography/data';

// This includes the @font-face declarations
import '@xo-union/typography/fonts';
```

Additionally, I added a mechanism for consumers to choose between using `data-urls` and `hosted-urls` for the fonts. I mostly started this as an experiment because I think there is some value to being able to have a different dev and production build, and this was a good opportunity to exercise the ability to do that. Besides that, I think there is some value to working with `data-urls` for development. Using data urls allow you to not rely on a network connection so it facilitates offline development. 

To switch between the modes, you can make the following changes to the webpack config file.

```js
const { DefinePlugin } = require('webpack');

module.exports = {
  // ommited
  plugins: [
    new DefinePlugin({
      'process.env.NODE_ENV': '"production"'
    })
  ]
}
```

pkelly:
@GeorgeTaveras1231: I'm very hesitate to include data-urls() for fonts. We definitely don't want people doing that in production because it will delay page rendering since CSS rendering blocks the initial render. Also we cannot use woff2 with data-urls() cause you cannot provide a fallback font for browsers that don't support woff2.

I'm wary of even doing it in dev because your dev env will now be different from prod. Page renders are different between environments. I like the idea of our various envs being as similar as possible.

That being said, it would be nice if you could work on Union with no internet. We can serve the fonts locally w/o leaning on data-urls() though.

pkelly:
@GeorgeTaveras1231: I'm not seeing the default TK styles for headers and whatnot (see attached).

screen shot 2017-02-15 at 4 59 58 pm

Can I add those back in? Or are you having concerns with including them globally?

GeorgeTaveras1231:
@pkelly as I mentioned earlier. The missing styles are due to [these `@import`s](https://github.com/xogroup/union/blob/master/pattern-library/src/typography/modules.scss#L7-L8) missing [here](https://github.com/xogroup/union/blob/master/pattern-library/src/typography/globals.scss)

pkelly:
Closing this PR out and creating a new one so I can squash all these changes into a clean commit on a new branch.
https://github.com/xogroup/union/tree/pk-add-type-styles-and-doc-page

Issue #45, opened 2017-02-08
Add doc page for foundations/typography


pkelly:
Migrate non static mock content from ~~https://xogroup.github.io/union/foundations/typography/~~ https://pkelly.github.io/union/foundations/typography/

- [x] Update specs for `<Snippet />`

pkelly:
Deployed.
http://xo-union-docs.s3-website-us-east-1.amazonaws.com/typography

PR #61, opened 2017-02-16
Add type doc page and update styles

pkelly:
Issue #45 

- Clean up typography.md
- Heading font face is tisa serif
- Use TK font scale for `<h*>`
- Set font-size on `<html>` so REMs are consistent across browsers
- Set`<p>` to $font-up-1
- Remove .article `h*` styles so headers are always rendered consistently
at the expense of some redundancy in spacer elements
- Update `<Snippet>` to be inline-able for cleaner doc pages

**Note**
This is a squash of another branch / PR. See old PR below for comment history
https://github.com/xogroup/union/pull/55


pattern-library/src/typography/globals.scss
```diff
@@ -4,5 +4,13 @@
 @require '@xo-union/typography/fonts';
 
 @import '~#/_shared/bootstrap-overrides';
+@import './variables';
 @import '~bootstrap/scss/type';
 
+html {
```
pkelly: CSS REMs are based on the html font-size. As far as I know modern browsers have a default of 16px, but I think we should be explicit in case a particular browser is different.

pattern-library/docs/articles/foundations/form-elements.md
```diff
@@ -21,18 +22,22 @@ $imports:
 
 <FormDemo />
 
-## Usage
+<div className={utils.mt4}></div>
```
pkelly: I agree that we should have a general wrapper with margin on it for the docs. I tried using `<section>` as the wrapper, however you can't have markdown inside html. Maybe there is a markdown thing we can do to create a wrapper.

pkelly: Maybe setting a margin on `<hr>`s for the docs. Markdown hr syntax is pretty nice.
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#horizontal-rule

Issue #48, opened 2017-02-10
Rename .tisa-* type+weight classes


GeorgeTaveras1231:
Currently, these classes are defined as followed:

```sass
.tisa-light {
  font-family: $font-family-serif;
  font-weight: 300;
}

.tisa-med {
  font-family: $font-family-serif;
  font-weight: 500;
}

.tisa-sans {
  font-family: $font-family-sans-serif;
  font-weight: 400;
}
```

I believe it's a good idea to avoid using the font name as the class names. 
I think better names would be: `.serif-light`, `.serif-med`, `.sans-serif`.

The reason behind this thought is that coupling the class name to the font name makes it more difficult to migrate to new fonts (which will likely be named differently). 

My ideal migration process in that scenario would be to change a variable value in one place (in this case, probably `$font-family-sans-serif` and `$font-family-serif`). Without feeling like we need to rename some classes and potentially change many references to these classes.

PR #85, opened 2017-03-23
Removes tightly coupled font class names

dmitrinesterenko:
[Addresses #48]


pattern-library/src/typography/modules/font-weights.scss
```diff
@@ -1,19 +1,15 @@
...
 }
 
-.tisa-med,
 .serif-med
 {
   font-family: $font-family-serif;
   font-weight: 500;
 }
 
-.tisa-sans,
```
GeorgeTaveras1231: Looks good but I think the references to these classes need to be removed from `typography.md` file

dmitrinesterenko: @GeorgeTaveras1231 looks that the typography.md file had been removed in an earlier PR. I merged master in here and that file is no longer present.

dmitrinesterenko:
@GeorgeTaveras1231 check this out again please.

pattern-library/src/typography/modules/font-weights.scss
GeorgeTaveras1231: The file is [here](https://github.com/xogroup/union/blob/master/pattern-library/docs/articles/pattern-library/core-components/typography.md)

PR #89, opened 2017-03-30
Add Section Header Component

Jacfem:


pattern-library/src/section-header/css/index.css
```diff
@@ -0,0 +1,7 @@
+@value black from '@xo-union/colors';
+
+.section-header {
+  color: black;
+  composes: lh-header tisa-sans font-up-1 from '@xo-union/typography';
```
GeorgeTaveras1231: Additionally, I believe this component should also have a blue bottom border.

pattern-library/src/section-header/index.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+export default function SectionHeader({ title = 'Hello world' }) {
+  return (
+    <div className={styles['section-header']}> {title} </div>
+  );
+}
+
+SectionHeader.propTypes = {
+  /**
+   * Name of Section Header
```
GeorgeTaveras1231: How about 'Title' of the Section Header?

packages/pattern-library/section-header/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@xo-union/section-header",
+  "version": "beta-1.0.0",
+  "main": "index.js",
+  "dependencies": {
+    "css-module-builder": "^1.0.0-beta.5",
+    "react": "^15.4.2"
```
GeorgeTaveras1231: This should include an `@xo-union/colors` dependency

packages/pattern-library/section-header/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@xo-union/section-header",
+  "version": "beta-1.0.0",
```
GeorgeTaveras1231: I think we should use a pre 1.0.0 since that's how the other packages are marked. I think `0.1.0` makes sense as an initial version

pattern-library/src/section-header/index.jsx
```diff
@@ -0,0 +1,19 @@
+import React, { PropTypes } from 'react';
+import '@xo-union/typography/fonts';
+import '@xo-union/typography/globals';
+import '@xo-union/typography/sizes';
+import '@xo-union/colors';
```
Jacfem: Oh, I think this was from when I developed locally in our application. I'll remove this, thanks!

pattern-library/src/section-header/index.jsx
Jacfem: Consistency 👍

packages/pattern-library/section-header/package.json
Jacfem: 👍

pattern-library/src/section-header/css/index.css
Jacfem: Thank you! Hmm, the width of this border will vary depending on the section one uses it in. Given that, would someone using this component customize the width in their own application?

Jacfem:
@GeorgeTaveras1231  Changes addressed, let me know if you see anything else! Thanks.

PR #94, opened 2017-04-05
Added grid breakpoint and grid container overrides

silviatc:
I could check lines 33, 34, and 43 and see if they need to be included

### DIFF
--- a/pattern-library/src/_shared/_bootstrap-overrides.scss
+++ b/pattern-library/src/_shared/_bootstrap-overrides.scss
@@ -29,7 +29,7 @@ $grid-breakpoints: (
   md: 824px,
   lg: 1004px,
   xl: 1320px
-) !default;
+);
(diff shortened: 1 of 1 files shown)


GeorgeTaveras1231:
@silviatc Those lines seem to be quick tests to make sure the grid 'makes sense'

silviatc:
@GeorgeTaveras1231 sounds good! do I merge or do you merge?

GeorgeTaveras1231:
@silviatc I'm doing a quick check to make sure the grid actually changes :)

GeorgeTaveras1231:
@silviatc there was a small bug that didn't allow the new grid values to be set. It seems (based on my observation) that once variables are set with the `!default` macro (as they are by tw bootstrap), Then using the `!default` macro again leaves the variable untouched. To fix the problem, I removed the `!default` macro and that fixed the grid.

GeorgeTaveras1231:
All good 👍

PR #129, opened 2017-05-02
Fixes copyright item text transform [Finishes #144421685]

smanwaring:
Fixes copyright item text transform [Finishes #144421685]

@GeorgeTaveras1231 Hey look what we did!


packages/pattern-library/header-nav/src/css/menu-items.scss
```diff
@@ -66,7 +66,7 @@
 
 .sub-menu-copy-right-item {
   composes: gray-6-fg from '@xo-union/colors/lib/utilities';
-  composes: sub-menu-item;
+  border-top: menu-border;
```
GeorgeTaveras1231: This is great guys. And it would be good to also have a utility class specifically for that property. Something like `menu-item-border` and re-use that in `sub-menu-copy-right-item` and `sub-menu-item`.

smanwaring:
@GeorgeTaveras1231 we updated the code.

packages/pattern-library/header-nav/src/css/menu-items.scss
```diff
@@ -66,7 +69,7 @@
 
 .sub-menu-copy-right-item {
   composes: gray-6-fg from '@xo-union/colors/lib/utilities';
-  composes: sub-menu-item;
+  composes: sub-menu-item-border;
   padding: .37rem .6rem;
   font-size: .6rem;
```
GeorgeTaveras1231: This font-size should probably come from a typography class. @smanwaring I will add a story for this.

PR #140, opened 2017-05-12
Footer social section

GeorgeTaveras1231:


smanwaring:
@GeorgeTaveras1231 

A few notes. Let me know if these would be better served as separate bugs/chores/stories/non-issues:

- Social icons stay focused when you navigate back to the page after clicking on one of them (they should not have a pseudo classes after a click. 

- Font on "Follow us" is a `serif-light` font, the original TK site has a sans-serif font. I looked into what fonts were available and it looks like the `sans-serif` font looks correct but there isn't a light version of it. Figured you may have come across this same thing which is why you landed on the lighter serif font.

- The `distributed packages` test fails when I run yarn test.

Otherwise I think code is clean and well organized!

GeorgeTaveras1231:
@smanwaring Thanks for the feedback. I will organize my answers using an ordered list to denote a response for each concern.

1. I believe this is something the browser does with all links, they are focused once you click on them. The links that open new windows just allow you to see this in action because the original document is still active after you click on the link.

2. I hadn't realized this, thanks for pointing this out. It's fine that this is not yet available as a typography class -- I will add a class named `sans-serif-light`.

3. The distributed packages tests should be passing but it depends on the state of the file system in your machine. I say that they should be passing because they are passing in circle ci. You should run `npm run clean && npm run setup` and try them again :) Make sure that you have committed everything because the clean command can be destructive at times.

smanwaring:
@GeorgeTaveras1231 

1. You're right! I tested this on the live TK site and the behavior is the same. For some reason I thought I saw it behave differently.

2. Made a note to write a story for this. Is it okay to merge this PR without the correct font?

3.  That worked. All tests passed!

GeorgeTaveras1231:
@smanwaring about the font, it seems this is a larger issue than I thought. I tried creating a new font class [here](https://github.com/xogroup/union/blob/master/packages/%40xo-union/tk-css-typography/src/modules/font-weights.scss) but learned that the new font doesn't work correctly. It seems that when defining `@font-face`'s [here](https://github.com/xogroup/union/blob/master/packages/%40xo-union/tk-css-typography/src/fonts.scss), we are limiting the `font-weight`'s that are allowed with each font. So not only does the new class not work but the existing class `.sans-serif-med` also, has no effect. We should confirm with the design group whether we intend to limit the number of `font-weight` options or we just haven't needed to add the new `@font-face` s yet

PR #152, opened 2017-05-25
Analytics for tools menu button

smanwaring:


smanwaring:
This branch is still a WIP since we need to finish documenting the new Hotspot component. I needed to push these up in order to start working on a different branch without these changes (since they had already been committed). If there is a work around for stashing already committed changes would love to know! Otherwise, apologies if this makes things messy. George and I will continue to work on this.

PR #154, opened 2017-05-31
Feature/contextual sub nav

smanwaring:


docs/articles/pattern-library/content-patterns/navigation/headerNavDemo.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
smanwaring: - [ ] Rename file to `demos/HeaderNavWithSubNavDemo.jsx`
- [ ] Rename component to HeaderNavWithSubNavDemo

PR #155, opened 2017-06-01
Add analytics track call for hamburger menu

smanwaring:
Additionally:
- Changed tools menu analytics track to read "Opened Tools Menu"
- Added analytics track call for open and close for hamburger menu


smanwaring:
@GeorgeTaveras1231 I noticed that there is an inconsistency with the name references of this icons. Sometimes it is referenced as "hamburger" and other times (per Story) it references "Primary". Likewise, the sub-menu button is sometimes called "sub-menu" but the analytics track reads "Opened Tools Menu".  I wanted to point this out since I was unsure of how to handle naming these in the test specs that I added. Take a look and let me know what you recommend for this and the other snippets of code that I added. Thanks!

Issue #194, opened 2017-07-14
`Textarea` components should have similar error functionality as the `Field` component


marioharper:
The `Textarea` components should have `state` and `validationMessage` props, just like the `Field` component. They should drive changing the display and styling depending on which state they are in (`neutral, valid, invalid`). This will ensure form error styling is consistent throughout The Knot.

PR #196, opened 2017-07-14
Add Textarea validation state and message.

marioharper:
## Description

Added validation state and message similar to how the `Field` component currenlty works / looks.
Also removed `TextareaWithoutLabel`. Now `Textarea` just adds a label if there is a `label` property passed to it. 

## Related Issue

#194 

## Motivation and Context

Keeps form validation consistent throughout The Knot. 

## Types of changes

- [x] Documentation
- [x] Breaking change (fix or feature that would cause existing functionality to change. you had to modify existing tests.)

## Checklist:

## Reviewer Checklist


packages/@xo-union/tk-component-fields/src/components/Textarea/index.jsx
```diff
@@ -1,18 +1,33 @@
...
-  label = labelize(name),
+  label,
+  state = 'neutral',
+  validationMessage,
   id = fieldId(name),
   ...props
 }) {
+  const inputClass = classMap[state];
+
+  if (!inputClass) {
```
GeorgeTaveras1231: should be named `textareaClass`.

packages/@xo-union/tk-component-fields/src/components/Textarea/index.jsx
```diff
@@ -1,18 +1,33 @@
...
+const classMap = {
+  neutral: FieldsCss.textarea,
+  invalid: FieldsCss.invalidTextarea,
+  valid: FieldsCss.validTextarea
+};
 
 export default function Textarea({
   name,
-  label = labelize(name),
+  label,
```
GeorgeTaveras1231: I think the preferred behavior would be for the label to be assumed when it is not provided. So

```jsx
<Textarea name="my-field" />
```

Should have a label of `"My field"`. This is helpful in most cases, and maintains parity with the current implementation of `<Field />`.

If a consumer wants to opt out of that behaviour, they can do:

```jsx
<Textarea name="my-field" label={false} />
```

packages/@xo-union/tk-component-fields/src/components/Textarea/index.jsx
```diff
@@ -1,18 +1,33 @@
...
+  state = 'neutral',
+  validationMessage,
   id = fieldId(name),
   ...props
 }) {
+  const inputClass = classMap[state];
```
GeorgeTaveras1231: We should have unit tests for this logic: Which class is applied based on the given `state`

marioharper: These unit tests cover that https://github.com/xogroup/union/pull/196/files#diff-d2374d2fb8d6cca58e5d25988675bea1R30.

marioharper:
@GeorgeTaveras1231 requested corrections made

GeorgeTaveras1231:
Thanks @marioharper 👍. We decided to merge this PR before #186 since it is much simpler. Sorry for the delay, and thanks again for the contribution.

GeorgeTaveras1231:
Release `@xo-union/tk-component-fields@1.0.0`

marioharper:
Awesome, thanks @GeorgeTaveras1231. Looks like I introduced some weirdness though... Looks like a z-index issue. Noticed this on the top demo at http://docs.union.theknot.com/pattern-library/core-components/form-elements. 

screen shot 2017-07-19 at 10 56 44 am

I'll open an issue to track this.

PR #202, opened 2017-07-20
Add FancyField component as content pattern

pkelly:
## Description

Add new content pattern, "Fancy field" that is a standalone input text box that can be inserted inline with text (see image below for example).

## Motivation and Context

This solves the problem of having a stylized input that can live within the flow of a text block. It differs from the standard inputs we already have because it isn't intended to be used on a form. The use case that motivated its design is a multi-step on-boarding flow.

## Types of changes

- [ ] Documentation

## Checklist:

## Reviewer Checklist

screen shot 2017-07-20 at 2 50 03 pm


jmsardina:
@pkelly It might make sense to have `/tk-component-fancy-field` become `/tk-component-fancy-fields` instead and create a `Field` component that is scoped to it.  This will allow us to encapsulate all of the fancy fields such as `Dropdown`, etc in a common directory.

pkelly:
@jmsardina: That makes sense. I didn't do that initially cause I was thinking of these as independent inputs, not part of a form set. However the dropdown is very much an inline element like the text field. I think that is a good plan 👍

packages/@xo-union/tk-component-fancy-field/src/index.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import styles from '@xo-union/tk-component-fancy-field/lib/css';
+import fontSizes from '@xo-union/tk-css-typography';
+
+export default function FancyField(props) {
+  return (
+    <div className={styles.wrapper}>
```
hsbacot: How can the

packages/@xo-union/tk-component-fancy-field/src/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+      <input
```
hsbacot: How can the consumer get the value from the ```input```?

packages/@xo-union/tk-component-fancy-field/src/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+        type="text"
+        value={props.value}
+        className={classNames(styles.field, props.fontClass, props.error && styles.invalid)}
+        placeholder={props.placeholder}
+        style={{ width: props.width, height: props.height }}
```
hsbacot: How can other attributes be applied to ```input```?

packages/@xo-union/tk-component-fancy-field/src/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+      />
+      {
+        props.error &&
+        <span className={styles.error}>{props.error}</span>
```
hsbacot: Is there a way to move the error message around?

hsbacot:
I have two major concerns with this pull request.

First: The component completely limits the control that the parent can have over it by removing access to the event handlers. This means the parent cannot access the ```value``` of the ```input```.
Second: This component is not flexible enough to be consumed because it restricts the user from passing any additional ```props``` that may be needed.

**I would mark this as needs changes and prevent merging immediately**

packages/@xo-union/tk-component-fancy-field/src/index.jsx
pkelly: Oversight on my part. I'll add in a handler for that.

packages/@xo-union/tk-component-fancy-field/src/index.jsx
pkelly: What kind of attributes are you thinking of?
Are there other props you think would be useful to configure the field?

packages/@xo-union/tk-component-fancy-field/src/index.jsx
pkelly: I was thinking the error msg styling would be rigid so that it looks consistent wherever it is used. You have an example of what that would look like?

packages/@xo-union/tk-component-fancy-field/src/index.jsx
GeorgeTaveras1231: I agree with @hsbacot. My opinion on this subject is that there are too many standard HTML attributes and we shouldn't limit their use when creating shared components. Thankfully, most of the time the flexibility can be added at a low cost. Just spread the rest of props on the *most important* element (most important is kind of a vague term but idk how else to describe it). However, this also requires that you *deconstruct* the props you need access to:

```jsx
function MyComponent({ firstProp, secondProp, ...restOfProps }) {
  return (
     <div firstProp={firstProp}>
        <span secondProp={secondProp> />
         <input {...restOfProps} /> {/* input being "the most important element" */}
     </div>
  );
}
```

GeorgeTaveras1231:
I really like the name fancy-field. Just felt the need to say that 😄

packages/@xo-union/tk-component-fancy-field/src/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+  );
+}
+
+FancyField.propTypes = {
+  value: PropTypes.string,
+  placeholder: PropTypes.string.isRequired,
+  /**
+  * Change the font size using a typography helper class
+  **/
+  fontClass: PropTypes.string,
```
GeorgeTaveras1231: I think instead of passing down the font class prop, consumers who want to change the font-size should do it via setting a font size on their form container. The input should just inherit it's container font-size, or use a relative font size via `em`s.

packages/@xo-union/tk-component-fancy-field/src/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+  width: PropTypes.string,
+  height: PropTypes.string,
```
GeorgeTaveras1231: It makes sense to me why the width may need to be specified, since this is an inline field. But I'm not sure about the height. To me, it seems like the height should be relative to the font-size of the container via the use of `em`s.

packages/@xo-union/tk-component-fancy-field/src/browser-spec.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import FancyField from '@xo-union/tk-component-fancy-field/src/index';
```
GeorgeTaveras1231: This import would be simpler if changed to `import FancyField from './'`;

packages/@xo-union/tk-component-fancy-field/src/index.jsx
hsbacot: It should be a composed export that relies on this input. This makes the assumption that every user of this needs error validation.

packages/@xo-union/tk-component-fancy-field/src/index.jsx
hsbacot: Agreed, I think this component should be little more than below if at all.
```javascript
const Input = ({ className, inputRef, ...rest }) =>
  (
    <input
      className={`${FieldStyles.field} ${className}`}
      ref={inputRef}
      type="text"
      {...rest}
    />
  );
```

packages/@xo-union/tk-component-fancy-field/src/index.jsx
hsbacot: At minimum is should expose ```className``` as a prop since it is the prop react provides and recommends using.

https://facebook.github.io/react/docs/dom-elements.html#classname

packages/@xo-union/tk-component-fancy-field/src/index.jsx
GeorgeTaveras1231: @hsbacot are you suggesting that consumers should (if they want errors in their fields) do:

```jsx
import { FancyField, FancyFieldError } from 'fancy-fields';

function MyComponent() {
  return (
    <FancyField {/* omitted */} />
    <FancyFieldError {/* omitted */} />
  );
}
```

packages/@xo-union/tk-component-fancy-field/src/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+import styles from '@xo-union/tk-component-fancy-field/lib/css';
+import fontSizes from '@xo-union/tk-css-typography';
+
+export default function FancyField(props) {
+  return (
+    <div className={styles.wrapper}>
+      <input
+        type="text"
+        value={props.value}
+        className={classNames(styles.field, props.fontClass, props.error && styles.invalid)}
```
GeorgeTaveras1231: This pattern for composing classes is not bad but it's inconsistent with how we have been composing classes in this project. We have been using the CSS Modules composition to combine classes that depend on each other. For example, here we would use a `invalid-field` class that composes `field` and adds the `invalid` styles. In our component, we would then have a simple conditional to chose which class to apply:

```jsx
if (props.error) {
  fieldClass = styles['invalid-field'];
}
```

I'm not completely opposed to adding `classnames` as a dependency but would rather use it everywhere or not use it at all. Also, I feel that in its most common use case, it is a lot like an array `filter` and `join`, for which we can just use the native methods.

packages/@xo-union/tk-component-fancy-field/src/index.jsx
hsbacot: @GeorgeTaveras1231 not quite. I meant that `<FancyFieldError />` should be composed around `<FancyField />` and exported on its own or by declaring it as a static on `FancyField`.

```jsx
import FancyField from './FancyField'
const FancyFieldError = props => {
  const { error, className } = props;
  return (
    <div className={styles.wrapper}>
      <FancyField
        type="text"
        className={`${classNames(styles.field, props.error && styles.invalid)}  ${className}`}
        {...rest}
      />
      {
        error &&
        <span className={styles.error}>{error}</span>
      }
    </div>
  )
}
```

This would allow us to export both `FancyField` and `FancyFieldError`

packages/@xo-union/tk-component-fancy-field/src/index.jsx
pkelly: That makes sense to me. Is there a reason `ref` can't be part of `...rest`?

GeorgeTaveras1231: Yea, because `ref` is not treated as a normal prop by React, you generally have to rename it if you want to pass it down.

packages/@xo-union/tk-component-fancy-field/src/index.jsx
GeorgeTaveras1231: @hsbacot I don't see how this is different functionally from my example.

I would say that the cost of including the error built into the component is so low, that I'd prefer to keep them together. If you don't want to render errors, just don't pass the `error` prop. 

However, if we decide that having the two separate is a must, I'd vouch for renaming the component to `FancyFieldWithError`.

Lastly, this is a slightly unrelated topic, but regarding the names of the props, @pkelly, can you try to mirror the names used in the other fields package? For example `errors` >> `validationMessage`.

pkelly: Creating two versions of the same component just so one doesn't have the error logic feels like unnecessary complexity. If we can accomplish the same HTML output through prop config and we don't yet need to reuse the error styling I would say we should keep things simple.

@GeorgeTaveras1231: I'll mirror the name conventions in the other fields package.

packages/@xo-union/tk-component-fancy-field/src/index.jsx
pkelly: I agree with both these points. Will make changes.

packages/@xo-union/tk-component-fancy-field/src/index.jsx
pkelly: I'll explore that. I think I may have added that to make vertical spacing easily controlled. I agree, that it would be ideal if you didn't have to specify that. Perhaps with adding spacing classes via classNames would allow the necessary control.

packages/@xo-union/tk-component-fancy-field/src/browser-spec.jsx
pkelly: Nice. Will change.

packages/@xo-union/tk-component-fancy-field/src/index.jsx
pkelly: I agree that generally `classnames` and array usage is equivalent. One benefit of using it here is that classnames flattens arrays for you. So the passed in className prop could be a string or an array. `classNames('a', ['b', 'c'])`

I also like that classNames ignores falsy values. `classNames(null, false, 'bar')`

Some people like the object syntax for conditional classes `classNames({ foo: true })` as well. Although I don't see much benefit of that when classes are referenced from a CSS file.

packages/@xo-union/tk-component-fancy-field/src/index.jsx
pkelly: Height was there to have the proper space between the text and bottom border. I changed it to use `padding-bottom` instead, which is looking good to me.

pkelly:
I've ported this over to @jmsardina's branch since she has created a fancy-fields package for this field and the dropdown she is working on.
https://github.com/xogroup/union/pull/220

Issue #213, opened 2017-07-21
Component Request - Multi-step form


peterkle:
For Guest List Manager onboarding sequence, I am building a multi-step form component which has no presentational aspect, but rather manages state. It will leverage the Fullscreen Component #201 and several other form-related components not yet build.

![screen shot 2017-07-21 at 6 14 05 pm](https://user-images.githubusercontent.com/486241/28484448-6a11ab3c-6e40-11e7-9e6e-f63dd18aaaa8.png)
![screen shot 2017-07-21 at 6 04 56 pm](https://user-images.githubusercontent.com/486241/28484288-7b0fee54-6e3f-11e7-8e18-71f0164ad08e.png)
![screen shot 2017-07-21 at 6 05 00 pm](https://user-images.githubusercontent.com/486241/28484286-7b0d9726-6e3f-11e7-9c20-ea86c7af94a7.png)

hsbacot:
Is there a reason we need to build this into a shared abstraction component? Managing the state of an application feels like it should be the applications job.

peterkle:
I expressed this concern with @GeorgeTaveras1231 regarding how much of the onboarding code to contribute, and he suggested to share state management and UI because it would reduce duplicative effort.

peterkle:
Discussed with @hsbacot after our meeting today. We decided it would be best in the long run to be  state mgmt agnostic in order be as flexible as possible. Closing this issue.

peterkle:
comments anyone? I closed the PR.

PR #215, opened 2017-07-21
Add a multi-step form component

peterkle:
## Description
This is a wrapper component that will do the state management for a
multi-step form. It gives children components the ability to advance to
the next step or go back.

## Related Issue
#213

## Motivation and Context
While this component doesn't have a UI, it is useful because it can be composed with other Union components, like the `<FullscreenModal />`, to create a sign up flow, for instance.

## Types of changes
- [ ] Documentation

## Checklist:

## Reviewer Checklist


hsbacot:
How does this benefit us in a way that using [react-router](https://reacttraining.com/react-router/web/guides/philosophy) wouldn't? We're already using it to handle the routing of the app, why not use it to handle the control flow as well? The proposed approach could introduce a lot of unnecessary complexity.

```jsx
<Router>
  <Multistep screens={{ screen1, screen2, screen3 }} />
</Router>
```
or
```jsx
<Router>
  <Route path="/screen1" component={Screen1} />
  <Route path="/screen2" component={Screen2} />
  <Route path="/screen3" component={Screen3} />
</Router>
```

They both handle switching screens in a conditional flow. One is already built and being used across our applications. I don't think we should reinvent the wheel if we don't have to.

Should we create documentation on how to accomplish this with ```react-router``` that is more specific to our use cases and include it here instead?

marioharper:
@hsbacot Personally, I'm in favor of having a `Multistep` form component if it is something that designers decide needs to be consistently styled and functioning throughout TK. If not, and there are many edge cases and differences in functionality, I'd lean towards documentation on how to accomplish multi step forms using react-router.

Issue #113, opened 2017-04-20
Update form field styling to match v2 in union-gui-kit


pkelly:
Update form field border, bg, and validation styling.
Use https://github.com/xogroup/union-gui-kit/blob/master/core_components.sketch (v2 section on forms page) as a reference

<img width="834" alt="screen shot 2017-04-20 at 1 27 31 pm" src="https://cloud.githubusercontent.com/assets/44373/25244095/5b53ed00-25cd-11e7-8287-bd8bd76904bb.png">

pkelly:
Closing this since I have not made progress in quite a while and comps are likely stale.

GeorgeTaveras1231:
Re-opening to allow interested contributors to pick this up

PR #219, opened 2017-07-25
Add RadioButton

### LINKED ISSUE #206: Component Request - RadioButton
New Component Request
===

## Description
There should be a `RadioButton` component in `@xo-union/tk-component-fields`. The Knot uses radio buttons on multiple forms.

### Vendor Request For Quote Form
screen shot 2017-07-21 at 10 33 04 am

**_Update_** 7/25/17
After discussion with two designers (Tyler & Rosina), we will be going with the design @peterkle commented with as the default radio button Union style : 

screen shot 2017-07-25 at 9 24 08 am

marioharper:
## Related Issue

closes #206 

## Types of changes

- [x] Documentation

## Checklist:

## Reviewer Checklist


marioharper:
@GeorgeTaveras1231 this PR is still in progress, but do you see anything in https://github.com/xogroup/union/pull/219/files#diff-65caee7278bd8b22f7c7e675638e3dac that can be optimized to use any of Unions helper classes/vars, etc?

hsbacot:
I think this is a solid implementation for now and this should hold unless we need more control over the radio options.  I don't think it's necessary yet but leaving an api that could work below for thought.

```jsx
<RadioGroup>
  <RadioOption value="Yes" {...otherProps} />
  <RadioOption value="No" {...otherProps} />
</RadioGroup>
```

marioharper:
@hsbacot I agree, I was just running with what you guys were doing on Acquisitions. I've seen it done both ways in different react radio button libraries. 

@GeorgeTaveras1231 any input on what you think would be the best API? I'm basically almost done with this PR, just finishing up a couple of unit tests.

We could run with this implementation and later implement `RadioGroup` and `RadioOption` and keep `RadioButtonList` support (which would at that point just be a different api on top of `RadioGroup` and `RadioOption`.

GeorgeTaveras1231:
@marioharper I agree with you guys about `<RadioGroup>` `<RadioOption>` being a better API. This is very similar to how we have implemented the `<Dropdown>` and `<DropdownItem>`. For the sake of consistency and flexibility, I'm tempted to suggest introducing the `RadioGroup` now, as long as it isn't too much additional work. I'm concerned that if we don't introduce it now, we will live with the `RadioList` for a while; which isn't bad because honestly, it covers the most common use case.  However, it could be one of those things where a more complex use case of the radio buttons is pushed back because developers don't want to refactor the `RadioList`.

As a side note: @marioharper I don't think these `Radio*` components belong in the `tk-component-fields` package. They seem to be more thematically in line with the `fancy-fields`.

marioharper:
@GeorgeTaveras1231 Yeah I'll put in the work to do `RadioGroup` and `RadioOption` now. And what is "fancy-fields"? What deems something fit for fields vs fancy-fields?

packages/@xo-union/tk-component-fields/src/components/RadioButtonItem/browser-spec.jsx
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    it('sets checked', () => {
+      const subject = getInput({ selected: true }).node;
+
+      expect(subject.checked).toBe(true);
+    });
+
+    it('calls handleChange when clicked', () => {
+      const clickHandlerSpy = jasmine.createSpy('clickHandlerSpy');
```
GeorgeTaveras1231: rename to `changeHandlerSpy`

packages/@xo-union/tk-component-fields/src/components/RadioButtonItem/index.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { PropTypes as T } from 'prop-types';
+import RadioButtonCss from '@xo-union/tk-component-fields/lib/css';
+
+const RadioButtonItem = (props) => {
+  const { label, value, name, selected, handleChange } = props;
+
+  const listItemClass = selected ?
+    RadioButtonCss.radioButtonListItemSelected : RadioButtonCss.radioButtonListItem;
```
GeorgeTaveras1231: Avoid using the camelized css modules. We are considering deprecating them.

packages/@xo-union/tk-component-fields/src/components/RadioButtonItem/index.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { PropTypes as T } from 'prop-types';
+import RadioButtonCss from '@xo-union/tk-component-fields/lib/css';
+
+const RadioButtonItem = (props) => {
+  const { label, value, name, selected, handleChange } = props;
+
+  const listItemClass = selected ?
```
GeorgeTaveras1231: Should rename `selected` to `checked` to mirror the `input`'s API

packages/@xo-union/tk-component-fields/src/components/RadioButtonItem/index.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { PropTypes as T } from 'prop-types';
```
GeorgeTaveras1231: For consistency, use the default export and name it PropTypes

packages/@xo-union/tk-component-fields/src/components/RadioButtonItem/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+      >
+        <div className={RadioButtonCss.radioButtonIcon} />
+        {label}
+      </label>
+    </li>
+  );
+};
+
+RadioButtonItem.propTypes = {
+  value: T.any,
```
GeorgeTaveras1231: How come the `value` prop isn't required but all other props are required?

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+  var-sp-down-1,
+  var-sp-down-2
+) from '@xo-union/tk-css-spacing';
+
+.radio-button-list {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+  overflow: visible;
+}
```
GeorgeTaveras1231: Use `m-0` and `p-0` from spacing.

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+
+.radio-button-list-item {
+  color: var-gray-4;
```
GeorgeTaveras1231: Use `composes: fg-gray-4 from '@xo-union/tk-css-colors'`

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+  } 
+}
+
+.radio-button {
+  position: absolute;
+  visibility: hidden;
+}
+
+.radio-button-label {
+  font-size: var-fs-up-1;
```
GeorgeTaveras1231: Use the immutable `font-up-1` from typography:

```css
composes: font-up-1 from '@xo-union/tk-css-typography'`;
```

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  flex-shrink: 0;
+  text-align: left;
+  margin: 0;
+}
+
+.radio-button-icon {
+  background-color: var-white;
```
GeorgeTaveras1231: Use `bg-white` from colors

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+  border: 1.3px solid var-safe-gray-7;
+  border-radius: 100%;
+  height: 23px;
+  width: 23px;
+  transition: border .25s linear;
+  margin-right: var-sp-down-1;
```
GeorgeTaveras1231: Use `mr-down-1` from spacing

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+  font-size: var-fs-up-1;
+  cursor: pointer;
+  -webkit-transition: all 0.25s linear;
+  color: var-gray-4;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  flex-shrink: 0;
+  text-align: left;
+  margin: 0;
```
GeorgeTaveras1231: Use the immutable `m-0` from spacing
```css
composes: m-0 from '@xo-union/tk-css-spacing';
```

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+
+.radio-button-icon {
+  background-color: var-white;
+  border: 1.3px solid var-safe-gray-7;
+  border-radius: 100%;
+  height: 23px;
+  width: 23px;
+  transition: border .25s linear;
+  margin-right: var-sp-down-1;
+  margin-top: -4px;
```
GeorgeTaveras1231: Curious why you needed to use a negative margin here

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+
+.radio-button-label {
+  font-size: var-fs-up-1;
+  cursor: pointer;
+  -webkit-transition: all 0.25s linear;
+  color: var-gray-4;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  flex-shrink: 0;
```
GeorgeTaveras1231: Curious why the `flex-shrink` is being specified without a `flex-basis` in the container.

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+  border: 1.3px solid var-safe-gray-7;
+  border-radius: 100%;
+  height: 23px;
+  width: 23px;
+  transition: border .25s linear;
+  margin-right: var-sp-down-1;
+  margin-top: -4px;
+  -webkit-transition: border .25s linear;
+  box-sizing: border-box;
+  flex-shrink: 0;
```
GeorgeTaveras1231: Curious why the flex-shrink is being specified without a flex-basis in the container.

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+  background-color: var-white;
+  border: 1.3px solid var-safe-gray-7;
+  border-radius: 100%;
+  height: 23px;
+  width: 23px;
+  transition: border .25s linear;
+  margin-right: var-sp-down-1;
+  margin-top: -4px;
+  -webkit-transition: border .25s linear;
+  box-sizing: border-box;
```
GeorgeTaveras1231: You shouldn't need the box-sizing here. We have a global rule that applies this to all elements:

```css
html {
  box-sizing: border-box;
}

* {
  box-sizing: inherit;
}
```

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+
+.radio-button {
+  position: absolute;
+  visibility: hidden;
+}
+
+.radio-button-label {
+  font-size: var-fs-up-1;
+  cursor: pointer;
+  -webkit-transition: all 0.25s linear;
```
GeorgeTaveras1231: Could this be more specific about which properties are being transitioned?

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  &:hover .radio-button-icon, .radio-button-icon {
+    border: 7px solid var-knot-blue-darker-30;
+  } 
+}
+
+.radio-button {
+  position: absolute;
+  visibility: hidden;
```
GeorgeTaveras1231: Is the combination of these two rules the same as `display: none`?

GeorgeTaveras1231:
@marioharper by `"fancy-fields"` I'm referring to the new types of fields. Also known as the Madlib style fields: https://github.com/xogroup/union/pull/202

marioharper:
@GeorgeTaveras1231 thanks for the feedback. Regarding the 'fancy-fields' the design team confirmed that this will be the default style of radio buttons. So I don't think this fits with fancy fields.

packages/@xo-union/tk-component-fields/src/components/RadioButtonItem/index.jsx
hsbacot: Any reason for deprecating them? I feel like they add value in the flow of `jsx` by providing an accessible dot property. It also allows for specific classes to be destructured from the object while importing. I'm not saying we should switch everything to it immediately, but I think it adds some value.

hsbacot:
I think it makes sense to refer the them as `fancy-*` because we will be moving towards northstar styles sooner rather than later.

marioharper:
What do you mean by northstar styles?

hsbacot:
The other option on the image in the issue.

tcmmct:
Based on the design unification meeting yesterday. The team aligned around these being the default radio buttons and not 'fancy.' They work for the fancy use case and the default use case

packages/@xo-union/tk-component-fields/src/components/RadioButtonList/index.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { PropTypes as T } from 'prop-types';
+import RadioButtonCss from '@xo-union/tk-component-fields/lib/css';
+import RadioButtonItem from '../RadioButtonItem';
+
+const RadioButtonList = ({ listItems, name, selecetedValue, handleChange }) =>
+  (<ul className={RadioButtonCss.radioButtonList}>
```
marioharper: what's the other option?

smanwaring: You can do `{RadioButtonCss['radio-button-list']}`.

packages/@xo-union/tk-component-fields/src/components/RadioButtonItem/index.jsx
GeorgeTaveras1231: @hsbacot we think it adds value too which is why we enabled them in the first place. The tradeoff is that when you add a lot of these css modules in a bundle, each module is literally taking up double their size because each key is duplicated.

so which is more important? syntactic sugar or bundle size?

We still haven't deprecated them, we are waiting on #182 to be finished so that we can measure the impact of removing that setting. If it is significant enough, we will probably give up the syntactic sugar.

Another thing for us to consider is to just use camel case for all classes which will work for CSS and allow the dot syntax. The tricky part is that we inherit a lot of classes from tw/bootstrap, which means we would need a way to delete the original classes and replace them for camelized versions. Another problem is that the grid module from tw/bootstrap comes with attribute rules that depend on the dashed classes:

```css
[class^="col-"] {
  /* omitted */
}
```

GeorgeTaveras1231:
@tcmmct I'm surprised because the other fields are gray, while this radio button is blue.

Unless you guys are using the fields from #113 to determine this.

tcmmct:
Yeah, @GeorgeTaveras1231 we do plan to update to a style like or similar to #113.

So, The 'fancy' style will be the default. It looks slightly off from the current form fields (which is fine). It will match after we update the other form fields.

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
marioharper: I'm not sure, pulled this over from Acquisitions, I'm no flex master. But removing `flex-shrink` makes the radio button get all squishy when the screen size is reduced.

marioharper:
@GeorgeTaveras1231 @hsbacot this is ready for full review now. API has been changed, tests have been made, and documentation has been updated.

docs/_/@components/PropTypesTable/index.jsx
```diff
@@ -3,19 +3,31 @@ import PropTypes from 'prop-types';
...
+// Removes keys from object by key name.
+const omit = (theObject, keys) => {
+  const filteredObject = Object.assign({}, theObject);
+
+  keys.forEach((prop) => {
+    delete filteredObject[prop];
+  });
+
+  return filteredObject;
+};
```
GeorgeTaveras1231: This is a candidate to be moved into `xojs` (our utility library). The library is not yet documented but it lives in `packages/xojs`. This function should be in `xojs/src/objects/omit`.

Additionally, all functions in `xojs` assume to be bound to a receiver. That just means that to be consistent, you should change `theObject` to be `this`. When used in this project, call it as:

```javascript
import omit from 'xojs/lib/objects/omit';

myObject::omit('key1', 'key2')

```

packages/@xo-union/tk-component-fields/src/components/RadioGroup/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+      key: index,
+      onChange,
+      ...child.props
+    });
+  }
+
+  return (
+    <div className={RadioButtonCss.radioGroup}>
+      {
+        children && children.map((c, i) => (renderChild(c, i)))
```
GeorgeTaveras1231: I believe you should use `React.Children.map` here

packages/@xo-union/tk-component-fields/src/components/RadioGroup/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+  function renderChild(child, index) {
+    return React.cloneElement(child, {
+      name,
+      index,
+      checked: child.props.value === value,
+      key: index,
+      onChange,
+      ...child.props
+    });
+  }
```
GeorgeTaveras1231: I like this approach. It seems much simpler than the implementation of `Dropdown` and `DropdownItem` which uses context.  👍

GeorgeTaveras1231: The only thing I would suggest is to move this function (`renderChild`) out of the component function. This helps avoid unnecessary memory allocation

packages/@xo-union/tk-component-fields/src/css/styles/radio-button.scss
```diff
@@ -0,0 +1,70 @@
+@value ( bg-white, var-safe-gray-7, var-knot-blue-darker-30, var-gray-4, var-black, var-white ) from '@xo-union/tk-css-colors';
+@value ( font-up-1 ) from '@xo-union/tk-css-typography';
+@value (
+  m-0,
+  p-0,
+  mr-down-1,
+  var-sp-down-1,
+  var-sp-down-2
+) from '@xo-union/tk-css-spacing';
```
GeorgeTaveras1231: I originally preferred to use `@value` for variables only. But I like how this helps organize the file. It feels more like a js module, with all imports at the top. I approve.

packages/@xo-union/tk-component-fields/src/components/RadioButton/index.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { PropTypes } from 'prop-types';
+import RadioButtonCss from '@xo-union/tk-component-fields/lib/css';
+
+const RadioButton = (props) => {
+  const { children, value, name, checked, onChange } = props;
```
GeorgeTaveras1231: Why not do the deconstructing in the arguments area? Makes sense since props is not being used for anything else.

docs/articles/pattern-library/core-components/form-elements/README.md
```diff
@@ -176,3 +178,23 @@ import { Field, Dropdown, DropdownItem, Textarea, FormTheme } from '@xo-union/tk
 <Demo>
   <Textarea name="optional-message" label="Optional Message" state="valid"/>
 </Demo>
+
+### RadioGroup Component
+
+<p>{RadioGroupMeta.description}</p>
+
+<Demo cssDependencies={{ bsGrid }}>
```
GeorgeTaveras1231: The cssDependencies is not necessary here

smanwaring:
You can ignore the issues on Code Climate (it was turned on by mistake). We've since turned it off, but for some reason the failure from the initial check is still showing up on this PR. It should go away next time the code is updated.

packages/@xo-union/tk-component-fields/src/components/RadioButton/index.jsx
marioharper: good callout 👍

marioharper:
@GeorgeTaveras1231 I addressed all comments except for migrating the `omit` method, I tried, but couldn't seem to get `xojs/lib/objects/omit` to resolve properly from within `docs/_/@components/PropTypesTable/index.jsx`

hsbacot:
This looks solid, nice work. It's a much stronger API now. 👍

GeorgeTaveras1231:
@marioharper great job on this!

packages/@xo-union/tk-component-fields/src/components/RadioButtonItem/index.jsx
hsbacot: @GeorgeTaveras1231 For clarification, the duplication is coming because we are camelCasing the classes that haven't been otherwise? It's not a prevention of us using camelCased class names?

GeorgeTaveras1231: @hsbacot the duplication is coming from the generated js. The generated js when using the camelize option is:

```javascript
module.exports = {
  'a-class': 'a-class--abc123 composed-class--abc123',
  'aClass': 'a-class--abc123 composed-class--abc123',
}
```

Issue #218, opened 2017-07-25
Component Request - Fancy Dropdown Component


jmsardina:
Guest Services is working on an on-boarding flow, which includes a dropdown menu that's stylistically different from the form dropdown.

<img width="216" alt="screen shot 2017-07-25 at 9 41 37 am" src="https://user-images.githubusercontent.com/10458903/28577871-951096d8-711d-11e7-80b6-6ffedfbedcfe.png">

marioharper:
@jmsardina this looks like the one on the Acquisition squads onboarding process as well. 
<img width="1070" alt="screen shot 2017-07-25 at 1 40 11 pm" src="https://user-images.githubusercontent.com/7682104/28588075-e44a5484-713e-11e7-89d3-fd5f948a9737.png">

I'll send you a link to that code in slack. (In case you haven't already implemented this)

Samueljoli:
@jmsardina Hey it seems like your in progress with this component. We should probably add the `in-progress` label, I can't seem to find any components to work on :(, you guys / gals are quick!

GeorgeTaveras1231:
Closing per #220

Issue #216, opened 2017-07-24
Component Request - Inline text field


pkelly:
Guest Services is working on an on-boarding flow that is essentially the same as the one acquisition has built.

A large inline-text field is a requirement for that flow (see below). Having this field be shared between the flows would be fantastic so that we can leverage the great work that has already been done by acquisition.

<img width="1008" alt="screen shot 2017-07-24 at 11 54 11 am" src="https://user-images.githubusercontent.com/44373/28532300-70ca6c2e-7067-11e7-8b36-de6176e096bd.png">

pkelly:
I've done work on this reflected in #202, however I'm holding off until we have our discussion to decide which teams are working on which components.

pkelly:
Based on the convo today I'm forging ahead!

pkelly:
This is now on PR #220 so that it can be packaged together with the Fancy Dropdown that @jmsardina is working on.

Samueljoli:
@pkelly ah man this one is taken as well :(, would it make sense to add the `in-progress` label seeing that this seems to be in pr-state

GeorgeTaveras1231:
Closing per #220

PR #220, opened 2017-07-25
Introduce Fancy Fields package

jmsardina:
## Description
This branch introduces the FancyFields package, which currently encapsulates the a 'fancy' `Field` and `Dropdown`.

## Related Issue
https://github.com/xogroup/union/issues/218
https://github.com/xogroup/union/issues/216

## Motivation and Context
This component has been requested by Guest Services and Acquisition.  Creating a new pattern rather than using the existing form fields will allow us to encapsulate all similar components.

## Types of changes

- [ ] Documentation

## Checklist:

## Reviewer Checklist


docs/articles/pattern-library/core-components/fancy-fields/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Import
+
+```javascript
+import { Dropdown } from '@xo-union/tk-component-fancy-fields';
+```
+
+### Dropdown Component
+
+<Demo>
```
jmsardina: This is resulting in two separate code snippets.  Is this normal?
screen shot 2017-07-25 at 5 12 12 pm

GeorgeTaveras1231: @jmsardina This may be due to a bug in the markdown renderer, it doesn't support multiline elements in jsx. Every element must be in a single line.

packages/@xo-union/tk-component-fancy-fields/src/css/index.css
```diff
@@ -0,0 +1,36 @@
...
+ * Class definitions are hashed in order to avoid polluting the global namespace.
+ * To use this class, the actual value must be imported into your code by requiring this
+ * module.
+ *
+ * For example:
+ *
+ *    const myCssModule = require('./path/to/module');
+ *
+ *    console.log(myCssModule.testClass); // Logs hashed class name
+ * */
```
GeorgeTaveras1231: Remove the template comments above

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
```diff
@@ -0,0 +1,87 @@
+@value var-knot-blue-darker-30, var-white from '@xo-union/tk-css-colors';
+@value ( caret-up, caret-down ) from '@xo-union/tk-component-icons/lib/legacy';
+
+.container {
+  color: var-knot-blue-darker-30;
```
GeorgeTaveras1231: Use `fg-knot-blue-darker-30` from colors:

```css
composes: fg-knot-blue-darker-30 from '@xo-union/tk-css-colors';
```

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
```diff
@@ -0,0 +1,87 @@
+@value var-knot-blue-darker-30, var-white from '@xo-union/tk-css-colors';
+@value ( caret-up, caret-down ) from '@xo-union/tk-component-icons/lib/legacy';
```
GeorgeTaveras1231: This module is going to be deprecated next week. All icons must now come from the `Icon` component. This does mean you may have to refactor the css and component to properly render the correct icon. Let us know if this ends up causing problems. Like I said, we are planning to remove all usages of this module next week so if anything we could take care of the refactor.

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
```diff
@@ -0,0 +1,87 @@
...
+
+.container {
+  color: var-knot-blue-darker-30;
+  position: relative;
+  text-align: center;
+  width: 150px;
```
GeorgeTaveras1231: As @pkelly did in his implementation of the fancy fields; it may make sense to accept the width of the field as a prop to the component. This will work for more use cases.

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    ~ * .dropdown-list {
+      display: block;
+    }
+  }
+}
+
+.dropdown-list {
+  background-color: var-white;
```
GeorgeTaveras1231: Use `bg-white` from colors

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
```diff
@@ -0,0 +1,87 @@
...
+  box-shadow: 1px 1px 4px 0 rgba(179,178,178,0.5);
+  composes: p-0 m-0 from '@xo-union/tk-css-spacing';
+  display: none;
+  list-style: none;
+  position: absolute;
+  width: 100%;
+
+  .dropdown-option {
```
GeorgeTaveras1231: This rule probably doesn't need to be nested inside of `dropdown-list` because there are no use cases for `dropdown-option` to be outside of `dropdown-list`. So to keep the rules as flat as possible, we can bring this rule to the root level.

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
```diff
@@ -0,0 +1,87 @@
...
+}
+
+.dropdown-field {
+  -webkit-touch-callout: none;
+  -webkit-user-select: none;
+  -khtml-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+
+  background-color: var-white;
```
GeorgeTaveras1231: Use `bg-white` from colors

hsbacot:
👍 for moving towards the API of:

```jsx
<Dropdown>
  <DropdownItem value=1>One</DropdownItem>
</Dropdown>
```

docs/articles/pattern-library/core-components/fancy-fields/README.md
jmsardina: 👍

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
jmsardina: Ahh --  The name 'legacy' should've been a red flag!

docs/articles/pattern-library/core-components/fancy-fields/demos/DropdownDemo.jsx
```diff
@@ -0,0 +1,40 @@
...
+  static propTypes = {
+    value: PropTypes.string,
+    options: PropTypes.arrayOf(PropTypes.string).isRequired
+  };
+
+  constructor(props, context) {
+    super(props, context);
+
+    this.state = { selected: this.props.value };
+    this.onSelectCallback = this.onSelectCallback.bind(this);
```
GeorgeTaveras1231: We have been using `@autobind` from `core-decorators` to accomplish this.

docs/articles/pattern-library/core-components/fancy-fields/demos/DropdownDemo.jsx
```diff
@@ -0,0 +1,40 @@
...
+class DropdownDemo extends React.Component {
+  static propTypes = {
+    value: PropTypes.string,
+    options: PropTypes.arrayOf(PropTypes.string).isRequired
+  };
+
+  constructor(props, context) {
+    super(props, context);
+
+    this.state = { selected: this.props.value };
```
GeorgeTaveras1231: This variable could be defined in the class scope which lets us remove the constructor function all together.

```jsx
class MyClass {
  state = { selected: this.props.value }
}
```

docs/articles/pattern-library/core-components/fancy-fields/demos/FancyFieldsDemo.jsx
```diff
@@ -0,0 +1,44 @@
...
+    super(props, context);
+
+    this.state = {
+      selectedOption: 'Designer'
+    };
+
+    this.handleDropdownSelect = this.handleDropdownSelect.bind(this);
+  }
+
+  handleDropdownSelect(e) {
```
GeorgeTaveras1231: We have been using `autobind` from `core-decorators` to accomplish this.

docs/articles/pattern-library/core-components/fancy-fields/demos/FancyFieldsDemo.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Dropdown, DropdownItem, Field } from '@xo-union/tk-component-fancy-fields';
+import bsGrid from '@xo-union/tk-css-grid';
+
+export default class FancyFormDemo extends Component {
+  constructor(props, context) {
+    super(props, context);
+
+    this.state = {
```
GeorgeTaveras1231: This variable could be defined in the class scope which lets us remove the constructor function all together.

```jsx
class MyClass {
  state = { /* Omitted */ }
}
```

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+    >
+      <div
+        tabIndex="0"
+        role="option"
+        data-label={props.label}
+        aria-selected={props.isSelected}
+        className={FancyFieldsCss.dropdownItem}
+      >
+        {props.label}
+      </div>
```
GeorgeTaveras1231: It seems like the label and the value of the `DropdownItem` are always the same. I think there should be a difference between what is rendered in the UI and what value is associated with the option.

```jsx
<DropdownItem value={1}>One</DropdownItem>
```

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
```diff
@@ -0,0 +1,48 @@
+.container {
+  composes: fg-knot-blue-darker-30 from '@xo-union/tk-css-colors';
+  display: inline-block;
+  margin-right: 20px;
```
GeorgeTaveras1231: Use `mr-up-1` from spacing

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-item.scss
```diff
@@ -0,0 +1,25 @@
...
+  cursor: pointer;
+  position: relative;
+
+  &:hover {
+    color: white;
+    background-color: var-knot-blue-darker-30;
+  }
+
+  &:not(:last-child) {
+    border-bottom: 1px solid #eeeeee;
```
GeorgeTaveras1231: This is using a non-standard color, is that intended?

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-item.scss
```diff
@@ -0,0 +1,25 @@
...
+  composes: fg-black from '@xo-union/tk-css-colors';
+
+  composes: p-down-1 from '@xo-union/tk-css-spacing';
+  composes: font-up-1 from '@xo-union/tk-css-typography';
+
+  cursor: pointer;
+  position: relative;
+
+  &:hover {
+    color: white;
```
GeorgeTaveras1231: Use `var-white` from colors

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-item.scss
```diff
@@ -0,0 +1,25 @@
...
+    background-color: var-knot-blue-darker-30;
+  }
+
+  &:not(:last-child) {
+    border-bottom: 1px solid #eeeeee;
+  }
+
+  &.is-selected {
+    color: white;
```
GeorgeTaveras1231: Use `var-white` from colors

packages/@xo-union/tk-component-fancy-fields/src/css/styles/field.scss
```diff
@@ -0,0 +1,52 @@
...
+
+ .field:focus::placeholder,
+ .field:focus::-webkit-input-placeholder {
+   color: transparent;
+ }
+
+ .field {
+   border-style: none;
+   padding-bottom: 5px;
+   border-bottom: solid 1px #c3c3c3;
```
GeorgeTaveras1231: This is using a non-standard color, is that intended?

packages/@xo-union/tk-component-fancy-fields/src/css/styles/field.scss
```diff
@@ -0,0 +1,52 @@
...
+ }
+
+ .field:focus {
+   border-bottom: solid 1px var-black;
+   outline: none;
+ }
+
+.invalid {
+  border-bottom-color: var-coral-darker-40;
+  margin-bottom: 25px;
```
GeorgeTaveras1231: use `mb-up-2` from  spacing

packages/@xo-union/tk-component-fancy-fields/src/css/styles/field.scss
```diff
@@ -0,0 +1,52 @@
...
+ }
+
+ .field:focus::placeholder,
+ .field:focus::-webkit-input-placeholder {
+   color: transparent;
+ }
+
+ .field {
+   border-style: none;
+   padding-bottom: 5px;
```
GeorgeTaveras1231: Use `pb-down-3` from spacing

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+          <ul className={FancyFieldsCss.dropdownList}>
+            { children }
+          </ul>
+        }
+      </div>
+    );
+  }
+}
+
+export default Dropdown;
```
GeorgeTaveras1231: Also, can you guys explain what is happening with the `width` in this component. Seems like some simple math but just want to know more about why it is necessary.

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  static defaultProps = {
+    style: { width: '150px', height: '51px' }
+  }
+
+  constructor(props, context) {
+    super(props, context);
+
+    this.state = { open: false };
```
hsbacot: This needs support the ability to be closed from the outside in case there are sibling `Dropdown` nodes and only one can be open.

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+    this.toggleMenu = this.toggleMenu.bind(this);
+    this.closeMenu = this.closeMenu.bind(this);
+  }
+
+  toggleMenu() {
+    this.setState({ open: !this.state.open });
```
hsbacot: Should be using the updater function provided by `setState` here since it's deriving the next state from the previous state.

https://facebook.github.io/react/docs/react-component.html#setstate

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
```diff
@@ -0,0 +1,45 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import FancyFieldsCss from '@xo-union/tk-component-fancy-fields/lib/css';
+
+const DropdownItem = (props) => {
+  return (
+    <li
+      className={`${FancyFieldsCss.dropdownItemBase} ${props.isSelected && FancyFieldsCss.isSelected}`}
+      data-label={props.label}
+      onMouseDown={props.onSelect}
```
hsbacot: Why `mouseDown` instead of `onClick`?

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+      style={props.style}
+    >
+      <div
+        tabIndex="0"
+        role="option"
+        data-label={props.label}
+        aria-selected={props.isSelected}
+        className={FancyFieldsCss.dropdownItem}
+      >
+        {props.label}
```
hsbacot: Should be `children`, not `label` here

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+const DropdownItem = (props) => {
+  return (
+    <li
+      className={`${FancyFieldsCss.dropdownItemBase} ${props.isSelected && FancyFieldsCss.isSelected}`}
+      data-label={props.label}
+      onMouseDown={props.onSelect}
+      style={props.style}
+    >
+      <div
```
hsbacot: Why do we have a `div` nested in a `li` that specifies it will render a string?

packages/@xo-union/tk-component-fancy-fields/src/components/Field/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+  inputRef,
+  validationMessage,
+  ...props
+  }) {
+  return (
+    <div className={styles.wrapper}>
+      <input
+        type="text"
+        className={classNames(styles.field, className, validationMessage && styles.invalid)}
+        style={{ width: width, height: height }}
```
hsbacot: How come we are setting styles inline instead of with css?

packages/@xo-union/tk-component-fancy-fields/src/components/Field/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+  );
+}
+
+Field.propTypes = {
+  /**
+  * Additional class or array of classes to be added to the <input>
+  **/
+  className: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.array
```
hsbacot: When would `className` be an array?

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
GeorgeTaveras1231: This may have been carried over from the implementation of the other `Dropdown` component. In that implementation, the `input:focus` pseudo selector is used to render the neighboring dropdown items container (Something like `.field:focus ~ .dropdown-items { display: block; }`) Which makes it so that the `click` event can never be fired off of the dropdown items because before the `click` event is fired, the `input` loses focus, which then hides the dropdown item elements. However, `mousedown` runs before `blur` of the focused element, which allowed the component to register a click on the dropdown item before hiding them.

GeorgeTaveras1231: If `focus` and `blur` aren't used to drive the hide/show behavior, I agree that `onClick` should be used instead of `onMouseDown`. However, I also think that we should use `focus` and `blur` to support tabbing through form controls.

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
GeorgeTaveras1231: I think this will be addressed naturally if `focus` and `blur` are used to drive the open/close behavior. I was going to suggest to accept a prop to declare the initial state but have a hard time thinking of a scenario where having a dropdown that is not focused but opened makes sense.

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
GeorgeTaveras1231: I agree with this though the other dropdown is not implemented this way.

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
GeorgeTaveras1231: I agree, this could probably be reduced to just the `li`

packages/@xo-union/tk-component-fancy-fields/src/components/Field/index.jsx
GeorgeTaveras1231: I think this is just asserting that an array is supported if you as a consumer chose to use an array. It happens to be supported because of the use of the `classnames` npm package.

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
hsbacot: @GeorgeTaveras1231 agreed, all good as long as some support for this is implemented.

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/browser-spec.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { shallow } from 'enzyme';
+import DropdownItem from './';
+
+describe('<DropdownItem />', function() {
+  beforeEach(function() {
+    this.onSelect = jasmine.createSpy();
```
hsbacot: This should be good to stay per the discussion in #235

packages/@xo-union/tk-component-fancy-fields/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@xo-union/tk-component-fancy-fields",
+  "version": "0.0.1",
```
jmsardina: Why are new packages released as 0.0.0?  It doesn't sound like it would make sense semantically.  Same with 0.0.1 🤔.
I would've thought it would default to 1.0.0 since it's the first version (?).

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+    */
+    children: PropTypes.node,
+    /**
+    * Dropdown styles
+    */
+    style: PropTypes.object
+  };
+
+  static defaultProps = {
+    style: { width: '150px', height: '51px' }
```
jmsardina: Maybe we don't need to be explicit about it and accept `style` prop instead.

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
```diff
@@ -0,0 +1,45 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import FancyFieldsCss from '@xo-union/tk-component-fancy-fields/lib/css';
+
+const DropdownItem = (props) => {
+  return (
+    <li
+      className={`${FancyFieldsCss.dropdownItemBase} ${props.isSelected && FancyFieldsCss.isSelected}`}
```
jmsardina: Ah right!

packages/@xo-union/tk-component-fancy-fields/src/components/Field/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+  height,
+  inputRef,
+  validationMessage,
+  ...props
+  }) {
+  return (
+    <div className={styles.wrapper}>
+      <input
+        type="text"
+        className={classNames(styles.field, className, validationMessage && styles.invalid)}
```
pkelly: I was hoping you'd budge on classNames, but I know when to pick my battles 😉
I'll replace with an array.

packages/@xo-union/tk-component-fancy-fields/src/components/Field/index.jsx
pkelly: Thought I had this taken care of. Might have gotten clobbered when I merged branches. Will take out.

packages/@xo-union/tk-component-fancy-fields/src/components/Field/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+        style={{ width: width, height: height }}
+        ref={inputRef}
+        {...props}
+      />
+      {
+        validationMessage &&
```
pkelly: I like to lean on the side of simplicity until we know for sure other states will be necessary for this component.
If someone needs that in the future they could make that change.

packages/@xo-union/tk-component-fancy-fields/src/components/Field/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+  **/
+  validationMessage: PropTypes.string,
+  /**
+  * Additional props for the <input>
+  */
+  props: PropTypes.object
+};
+
+Field.defaultProps = {
+  width: '275px'
```
pkelly: Agreed. 1 more px makes it a multiple and 4 and shouldn't have a major design impact. I'll test it out to make sure though.

packages/@xo-union/tk-component-fancy-fields/src/components/Field/index.jsx
pkelly: Since the width is a passed in prop, it needs to be a style, right?

Sidenote: I'm removing the height param.

packages/@xo-union/tk-component-fancy-fields/src/components/Field/index.jsx
pkelly: Yep. I find it useful to apply a few classes inline. Common example being setting margin and padding. Great for things that make sense in the context of elements around them.

Alternative approach would be to always reference a class in the css file and use `composes`. For simple cases like spacing mentioned above this seems like overkill to me. 

Like @GeorgeTaveras1231 said though, this is really easy due to the use of `classnames` which I'm removing since it doesn't have a precedent yet.

I'll change this to a string and keep it simple for now.

packages/@xo-union/tk-component-fancy-fields/package.json
GeorgeTaveras1231: @jmsardina those are good points. Our preferred method is to use `0.1.0` for the first release. The reason why we need to start packages at `0.0.0` is that our publishing system always asks us to bump the version, even on the first release of a package. So starting at `0.0.0` allows us to bump to `0.1.0` (or any other one if we decide to do so) when releasing

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
jmsardina: You're right, there should be a separation!

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-item.scss
jmsardina: This was taken from our design specs, but I've replaced it with a color that consists with Union.

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
jmsardina: As @GeorgeTaveras1231 mentioned, `mouseDown` is used here to allow the selection to be registered before 'blurring'.  Otherwise, the natural order of events will result in the blur event happening before successfully setting the selected value.

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
jmsardina: That makes sense!  This will give the client control over the contents of this components, allowing them to pass in icons if needed.

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+      data-label={props.label}
```
GeorgeTaveras1231: Yup. This method of managing the values can also be problematic if consumers insert other html elements inside of the dropdown (Such as an icon). If the user clicks on the icon, the `data-label` value is lost. 

You can do two things to address this: 

You can use `event.currentTarget` instead of `event.target`. As long as the event handler is on the element with the `data-label` attribute, `event.currentTarget.dataset.label` will always have the right value, even if the event is triggered on another child element.

Secondly, you can avoid the use `data-` attributes altogether. `data-` attributes are most useful when losing scope of the element so storing some data that will be accessible later (maybe when querying the dom for the element) is useful. In here, the value is most useful to the caller of the dropdown. The caller will probably expect the selected value as an argument to `onSelect`, so as long as `props.value` (or `props.label`) is passed to `onSelect` I think you are good.

packages/@xo-union/tk-component-fancy-fields/src/css/styles/field.scss
```diff
@@ -0,0 +1,52 @@
+@value (
+  var-coral,
+  var-safe-gray-5,
```
smanwaring: Because we merged ( #224 ) `var-safe-gray-5` will need to be updated to `var-gray-5`. Likewise, `var-safe-gray-7` will need to be updated to `var-gray-7`.

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
jmsardina: I removed the nested `div`, so this is no longer an issue.

packages/@xo-union/tk-component-fancy-fields/src/components/DropdownItem/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+      onMouseDown={props.onSelect}
+      style={props.style}
```
jmsardina: It's not.  I've removed it and will pass in the expanded props instead.

docs/src/components/HeaderNav/CoreComponentsSubNav/index.jsx
```diff
@@ -17,6 +17,7 @@ export default class CoreComponentsSubNav extends Component {
         <NavItem to="/pattern-library/core-components/colors"> Colors </NavItem>
         <NavItem to="/pattern-library/core-components/buttons"> Buttons </NavItem>
         <NavItem to="/pattern-library/core-components/form-elements"> Form Elements </NavItem>
+        <NavItem to="/pattern-library/core-components/fancy-fields"> Fancy Fields </NavItem>
```
jmsardina: FancyFields is a separate package.  Could that create confusion?
I can go ahead and move this to the Form Elements documentation.

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
jmsardina: Width and height have been removed from the props.

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+    const { children, onSelect, value, label, ...otherProps } = this.props;
+
+    return (
+      <div
+        tabIndex="0"
+        className={FancyFieldsCss.container}
+        onBlur={this.closeMenu}
+        {...otherProps}
+      >
+        <div
```
GeorgeTaveras1231: I'm concerned with removing the `input` element because it won't allow a form using this component to fallback to native form submission. In general, I prefer this progressive approach for building custom form controls. This is not to say that this is an agreed upon Union standard, but it is probably worth discussing.

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+        { this.state.open &&
+          <ul className={FancyFieldsCss.dropdownList}>
+            {
+              React.Children.map(children, (child) => {
+                return React.cloneElement(child, {
+                  onSelect: this.props.onSelect,
+                  isSelected: this.props.value === child.props.value,
+                  ...child.props
+                })
+              }, this)
```
GeorgeTaveras1231: Nice! Only thing is the `this` is not needed at the end because you are using arrow functions which maintain the `this` of the scope.

docs/src/components/HeaderNav/CoreComponentsSubNav/index.jsx
GeorgeTaveras1231: You can probably still leave them in their own page, but leave a link in the "Form Elements" page that points to the fancy fields documentation

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
```diff
@@ -0,0 +1,50 @@
...
+  box-shadow: 1px 1px 4px 0 rgba(179,178,178,0.5);
+  list-style: none;
+  position: absolute;
+  width: 100%;
+}
+
+.caret {
+  cursor: pointer;
+  display: inline-block;
+  margin-left: 10px;
```
GeorgeTaveras1231: This should use `ml-down-1` or `ml-down-2` instead of a custom margin

packages/@xo-union/tk-component-fancy-fields/src/css/styles/dropdown-field.scss
```diff
@@ -0,0 +1,50 @@
...
+  -khtml-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+
+  border: none;
+  color: inherit;
+  cursor: pointer;
+  display: inline-block;
+  height: 100%;
+  margin-left: -21px;
```
GeorgeTaveras1231: Concerned about committing this non-standard negative spacing, as it would be the first in the project. Is there another way to accomplish the same that doesn't require this?

packages/@xo-union/tk-component-fancy-fields/src/css/styles/field.scss
```diff
@@ -0,0 +1,55 @@
...
+.field:focus {
+  border-bottom: solid 1px var-black;
+  outline: none;
+}
+
+.invalid {
+  composes: mb-up-2 from '@xo-union/tk-css-spacing';
+
+  border-bottom-color: var-coral-darker-40;
+  margin-bottom: 25px;
```
GeorgeTaveras1231: This should use `mb-up-2` instead of a custom margin.

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/browser-spec.jsx
```diff
@@ -0,0 +1,92 @@
...
+  beforeEach(function() {
+    this.onSelect = jasmine.createSpy();
+
+    this.wrapper = mount(
+      <Dropdown value="false" label="Declined" onSelect={this.onSelect}>
+        <DropdownItem label="Accepted" value="true" />
+        <DropdownItem label="Declined" value="false" />
+        <DropdownItem label="No Response" value="null" />
+      </Dropdown>
+    ).mount();
```
GeorgeTaveras1231: Is this separate `.mount()` call intended? Does this do a double mount?

docs/articles/pattern-library/core-components/fancy-fields/README.md
```diff
@@ -0,0 +1,102 @@
+---
+$imports:
+  'TkComponentFancyFields': '@xo-union/tk-component-fancy-fields'
```
GeorgeTaveras1231: This import is unused

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/index.jsx
jmsardina: Ahh... I understand the concern.  I'll go replace this div with an `input` element instead.

packages/@xo-union/tk-component-fancy-fields/src/components/Dropdown/browser-spec.jsx
jmsardina: I initially was using `.shallow`, and needed to call it twice to render the child elements.  But I only need one mount.

smanwaring:
@jmsardina this is ready to go besides the failing tests. Can you take a look at those when you get the chance?

jmsardina:
@smanwaring Yes! I just finished addressing the linter errors and updated the package.json to include the missing dependencies.

jmsardina:
@smanwaring  I updated the branch to set the required dependencies on `package.json` and address the linter errors.

smanwaring:
@jmsardina 👍 sweet!

PR #227, opened 2017-07-27
Fix the flash when server side rendered, remove all caps default butt…

RohitRox:
When rendered server side, the component flashes when it adjusts the height of the container. This PR fixes that issue with alternate height adjustment solution. It also resets the default button text case to be regular, Madeline mentioned we no longer use all caps for buttons. I have also added a nice little css based fade in for the read more button, as button gets rendered last after height adjustment it felt like blink if displayed without fade.

## Types of changes

- [ ] Documentation
- [x] Bug fix (non-breaking change which fixes an issue. you didn't modify existing tests)

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/xogroup/union/blob/master/.github/CONTRIBUTING.md) document.
- [ ] I have updated the documentation as needed.

## Reviewer Checklist


packages/@xo-union/tk-component-more-or-less/src/index.jsx
```diff
@@ -13,18 +13,25 @@ export default class MoreOrLess extends Component {
...
       truncatable: false,
-      truncated: false,
+      truncated: true,
     };
   }
 
   /* eslint-disable react/no-did-mount-set-state */
   componentDidMount() {
+    this.contentElement.style.height = 'auto';
+
```
smanwaring: This is now a bit bloated. Is there any way to split these up into separate methods on this component and call them instead?

packages/@xo-union/tk-component-more-or-less/src/index.jsx
```diff
@@ -84,9 +91,9 @@ export default class MoreOrLess extends Component {
 
   render() {
     return (
-      <div>
+      <div className={styles.wrapper} ref={(el) => { this.contentElementWrap = el; }}>
         <div
-          className={styles.container}
+          className={styles.contentContainer}
```
smanwaring: camelCase class imports might be deprecated. Use ['content-container'] instead.

packages/@xo-union/tk-component-more-or-less/src/utilities/truncator.js
```diff
@@ -0,0 +1,28 @@
...
+    this.element.style.height = `${allowedHeight}px`;
+    this.element.style.overflow = 'hidden';
+  }
+
+  untruncate() {
+    this.element.style.height = 'auto';
+    this.element.style.overflow = 'visible';
+  }
+
+}
```
GeorgeTaveras1231: There should be tests for this class

packages/@xo-union/tk-component-more-or-less/src/index.jsx
```diff
@@ -119,6 +111,7 @@ MoreOrLess.propTypes = {
 
 MoreOrLess.defaultProps = {
   lines: 4,
-  readMoreText: 'READ MORE',
-  readLessText: 'READ LESS'
+  readMoreText: 'Read More',
+  readLessText: 'Read Less'
 };
+
```
GeorgeTaveras1231: @RohitRox This component seems a bit complicated and I'm concerned about future changes becoming hard to implement or test.

I have a simple recommendation but tell me what you think.

First, I think a reason why this component comes off as complicated is because there are side-effects modifying elements that are mounted in a lot of unexpected places. Some of these side-effects are happening on `componentDidMount` and are now happening in the `Truncator` class.

I think it would be much simpler to follow if you only used state to modify the elements. So instead of doing `this.element.style[styleProperty] = newValue`, use `.setState()` to update these values, and then assign the values in the `render` method. I think this will make the flow of changes a lot easier to follow.

packages/@xo-union/tk-component-more-or-less/src/utilities/truncator.js
```diff
@@ -0,0 +1,28 @@
+import { g } from 'xojs/lib/runtime/getGlobal';
+
+export default class Truncator {
+  constructor(element, lines) {
+    this.element = element;
+    this.lines = lines;
+    this.element.style.height = 'auto';
```
GeorgeTaveras1231: Concerned about this side-effect in the constructor. I believe it is a best practice for constructors to be pure as it aligns the flow of your program with what most developers expect. 

Put simply, developer's wouldn't expect constructors to modify their arguments, just to configure the object being created.

packages/@xo-union/tk-component-more-or-less/src/css/index.scss
```diff
@@ -1,37 +1,48 @@
...
   line-height: 27px;
   position: relative;
+  height: 107px;
+  overflow: hidden;
+  animation: fadein 0.5s ease-in-out;
+}
+
+@keyframes fadein {
+    from { opacity: 0; }
+    to   { opacity: 1; }
```
GeorgeTaveras1231: This should be refactored to use `transitions` instead. There is an outstanding issue with `a-css-loader` that makes key-frame definitions global, which they shouldn't be.

I opened #245

RohitRox:
@GeorgeTaveras1231 Hi, I totally understand that it is bad practice doing things like that in `componentDidMount`, and it is true that this thing looks more complicated than it should be. It is this way because it had to be designed to support server side render. All the heights, getComputedStyle and minHeight manipulation are there to fix the ui issues when server-side rendered. I can refactor the code to move those things in a separated methods (has to be called in componentDidMount though), wanted to let you know why those weird manipulations

GeorgeTaveras1231:
@RohitRox @gsheppard I know changes are planned for this component but I'd rather merge this as it is a better presentation of the code for the current version of the component, in case the re-implementation happens a long time from now.

Issue #204, opened 2017-07-21
Component Request - Generic Typeahead


marioharper:
New Component Request
===

## Description
There should be a **generic** `Typeahead` component in `@xo-union/tk-component-fields`. The Knot uses typeaheads in 2 instances that I know of (Vendor Typeahead, Location Typeahead). This is not a request for those instances, it is a request for a generic component that will aid in the creation of those instances.

### Location Typeahead
![screen shot 2017-07-21 at 10 15 36 am](https://user-images.githubusercontent.com/7682104/28470072-21b41a7e-6dfe-11e7-81ac-d6dfc3d3b417.png)

### Vendor Typeahead
![screen shot 2017-07-21 at 10 16 07 am](https://user-images.githubusercontent.com/7682104/28470076-265e97d4-6dfe-11e7-861d-5f5043cb5691.png)

GeorgeTaveras1231:
Timeline: Thurday 27th, July

hsbacot:
PR on track for tomorrow. Sorry about the delay.

GeorgeTaveras1231:
@marioharper @hsbacot I noticed that with the components from #231 we are still missing a lot of the integration steps between the field and the list of results. Are there plans from anyone to implement the specific components? Example: `LocationTypeahead`.

hsbacot:
@GeorgeTaveras1231 There are no plans to support fetching data, that functionality should be composed by the consumer. The consumer should be able to use this list for any type of suggestions provided from an api or otherwise. They would manage the state of the input, the state of the requests, and be able to compose a typeahead by supplying the data to `SuggestionsList`

marioharper:
I agree with @hsbacot, this should be generic for flexibility. We could author a `LocationTypeahead` plugin/module that utilizes this component though. I think Union itself should stay generic and stay as the building blocks, however, we could author separate plugins/modules that are created using Union that are used specifically by XO Group. This modules do not need to be publicly facing like Union is. They would be internal implementations. Any thoughts?

GeorgeTaveras1231:
@marioharper @hsbacot I'm not suggesting to dump the generic implementation for the specific one, they are both useful to have. @marioharper I agree and that's exactly what I was getting at: We will have the generic implementation available after #231 is merged but it would be useful to have a package for the specific use cases of this component. As far as where it should live, I would be ok with it living elsewhere if there were some benefits to doing that. To me it would seem like separation for the sake of separation. 

I would much rather keep that package in this repository because:

[...]

PR #231, opened 2017-07-28
SuggestionList Component

hsbacot:
This adds a SuggestionList & SuggestionListItem to the fields
package availaible in union

## Related Issue

#204 

## Types of changes

- [ ] Documentation

## Checklist:

## Reviewer Checklist

- [x] Is change documented in a release draft? (Only if there are plans to merge PR)


packages/@xo-union/tk-component-fields/src/components/SuggestionList/SuggestionListItem.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import noop from 'lodash/noop';
+import css from '@xo-union/tk-component-fields/lib/css';
+
+const SuggestionListItem = ({ id, description, onClick }) =>
```
smanwaring: You should `children` instead of `description` to make the component's API more robust. 

For example, the following use case is pretty 'ugly'.
```jsx
<SuggestionListItem description={<DescriptionWrapper><a>link</a> Description</DescriptionWrapper>} />
```

In comparison to:
```jsx
<SuggestionListItem>
  <DescriptionWrapper>
    <a>link</a> Description
  </DescriptionWrapper>
</SuggestionListItem>
```

smanwaring: Component should accept `...props`

packages/@xo-union/tk-component-fields/src/components/SuggestionList/index.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import css from '@xo-union/tk-component-fields/lib/css';
+import SuggestionListItem from './SuggestionListItem';
+
+const SuggestionList = ({ suggestions, onClick, className, children }) =>
+  (<ul className={`${css['suggestions-list']} ${className}`}>
```
smanwaring: `<ul>` should have a `role` that can serve as a container for the list-items which have `role="menuitem"`

packages/@xo-union/tk-component-fields/src/components/SuggestionList/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+    {suggestions.map(suggestion =>
+      (<SuggestionListItem
+        key={suggestion.id}
+        description={suggestion.description}
+        onClick={onClick}
+      />)
+    )}
+    {children}
```
smanwaring: Is there a reason this supports two interfaces to populate the list? It seems like it could work with just the `children` implementation (no need for a map and passing in a `key`). I feel like using `children` is more flexible and consistent with other implementations of rendering a list within the project.

```jsx
<SuggestionList>
  <SuggestionListItem className={} onClick={} {/* omitted */}> Option1 </SuggestionListItem>
</SuggestionList>
```

smanwaring: Also, the map implementation assumes that all list-items will have the same onClick behavior. Is that intended?

packages/@xo-union/tk-component-fields/src/css/index.scss
```diff
@@ -12,21 +12,18 @@
...
 ) from './styles/radio-button';
 
-@value (
-  invalid-textarea,
-  valid-textarea,
-  textarea-requirements
-) from './styles/textarea';
+@value (invalid-textarea,
+ valid-textarea,
+ textarea-requirements) from './styles/textarea';
```
smanwaring: Any reason to change the style of these variables?

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
```diff
@@ -0,0 +1,47 @@
...
+  position: absolute;
+  width: 275px;
+  z-index: 10;
+  height: auto;
+}
+
+.suggestions-item {
+  composes: m-0;
+
+  padding: 0 8px;
```
smanwaring: Use `py-0 px-down-2` from spacing

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
```diff
@@ -0,0 +1,47 @@
...
+  list-style: none;
+  text-align: left;
+  font-size: 20px;
```
smanwaring: Use `font-up-2` from typography

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
```diff
@@ -0,0 +1,47 @@
...
+  line-height: 51px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+
+  &:hover {
+    cursor: pointer;
+    color: var-knot-blue-darker-30;
+    background-color: rgba(62, 169, 218, 0.03);
```
smanwaring: This is a non-standard color. Is that intended?

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
```diff
@@ -0,0 +1,47 @@
+@value (bg-white, var-black, var-knot-blue-darker-30, bg-safe-gray-7) from '@xo-union/tk-css-colors';
+@value (m-0, p-0) from '@xo-union/tk-css-spacing';
+
+.suggestions-list {
+  composes: m-0 p-0 bg-white;
+  box-shadow: 1px 1px 4px 0 rgba(179, 177, 177, 0.5);
+  position: absolute;
+  width: 275px;
```
smanwaring: Does it make sense to have one hard-coded width for all instances of this component?

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
```diff
@@ -0,0 +1,47 @@
...
+  &:hover {
+    cursor: pointer;
+    color: var-knot-blue-darker-30;
+    background-color: rgba(62, 169, 218, 0.03);
+  }
+}
+
+.suggestions-footer {
+  composes: bg-safe-gray-7;
+  font-size: 13.5px;
```
smanwaring: Use `font-down-1` or `font-down-2` from typography

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
```diff
@@ -0,0 +1,47 @@
...
+  width: 100%;
+  height: 68.7px;
```
smanwaring: Does it make sense to hardcode this height?

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
```diff
@@ -0,0 +1,47 @@
...
+  white-space: nowrap;
+
+  &:hover {
+    cursor: pointer;
+    color: var-knot-blue-darker-30;
+    background-color: rgba(62, 169, 218, 0.03);
+  }
+}
+
+.suggestions-footer {
```
smanwaring: This class is not used.

packages/@xo-union/tk-component-fields/src/components/SuggestionList/SuggestionListItem.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import noop from 'lodash/noop';
```
smanwaring: This is a good candidate for xojs. We think it makes sense to move it there.

hsbacot: This is an external package. How would it make more sense in `xojs`?

packages/@xo-union/tk-component-fields/src/components/SuggestionList/index.jsx
hsbacot: @smanwaring  `children` could be used as a second way to populate the list, however the primary use case of `children` is to allow for a footer, or any other content to be passed through and fill to the size of the suggestion list.

I don't believe that the proposed api is beneficial in this case for a few reasons. `SuggestionList`'s primary use case is to map over dynamically generated data that is not explicitely controlled by the teams consuming it.

The others that follow this api, `Dropdown` & `RadioGroup`, have primary use cases if working with statically defined data.

I'm for supporting it if we collectively feel that this is a better approach for these use cases. It will put the work of mapping the items on the client when handled exclusively this way.
```jsx
const suggestions = [
  { id: 'suggestion 1', description: 'Suggestion One' },
  { id: 'suggestion 2', description: 'Suggestion Two' },
]
const suggestionItems = suggestions.map(suggestion =>
  (<SuggestionListItem
    key={suggestion.id}
    description={suggestion.description}
 />)
<SuggestionList onItemClick={onItemClick}>
  {suggestionItems}
</SuggestionList>
```
vs.
```jsx
const suggestions = [
  { id: 'suggestion 1', description: 'Suggestion One' },
  { id: 'suggestion 2', description: 'Suggestion Two' },
]
<SuggestionsList suggestions={suggestions} />
```

hsbacot: Yes, the mapping makes that assumption but it also provides the ability to support a non standard `onClick`.

Also, I agree that the naming of props here could be stronger.

packages/@xo-union/tk-component-fields/src/css/index.scss
hsbacot: No, I must've accidentally saved when viewing the file and my auto-formatter picked it up.

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
hsbacot: Yes, we need the alpha channel for hover.

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
hsbacot: Good catch, this shouldn't have been included.

packages/@xo-union/tk-component-fields/src/components/SuggestionList/SuggestionListItem.jsx
```diff
@@ -0,0 +1,21 @@
...
+import css from '@xo-union/tk-component-fields/lib/css';
+
+const SuggestionListItem = ({ id, description, onClick }) =>
+  (<li role="menuitem" className={css['suggestions-item']} key={id} onClick={onClick}>
```
hsbacot: Yes, this is meant to be used as a suggestion list for a typeahead's autocomplete. Users should be able to select suggestions from this list.

hsbacot:
@smanwaring It's in the `fields` package because it is meant to be paired with an `input` to create a typeahead suggestion list.

smanwaring:
@hsbacot gotcha. Perhaps we should add those details to the documentation for it and possibly show an example (if it makes sense to do so)?

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
```diff
@@ -0,0 +1,29 @@
+@value (bg-white, var-knot-blue-darker-30, bg-safe-gray-7) from '@xo-union/tk-css-colors';
```
smanwaring: Because we merged ( #224) you will need to change `safe-gray-7` to `gray-7`.

packages/@xo-union/tk-component-fields/src/css/styles/suggestion-list.scss
GeorgeTaveras1231: There is a high possibility that this can be replaced by one of the standard color. I'm also assuming that the alpha is not there to make the items translucent but to achieve a particular shade.

packages/@xo-union/tk-component-fields/src/components/SuggestionList/SuggestionListItem.jsx
GeorgeTaveras1231: I actually like the idea of adding such utilities (even if they come from an external package) to xojs. The idea, (though it doesn't apply to this function in particular) is that if everyone opts into using `xojs` as their utility library, we can centralize optimizations (by outsourcing or writing from scratch, functions that are faster).  However, we haven't really advertised the use of xojs but have kept it as a standard within the union project.

GeorgeTaveras1231: Here is a good example of an outsourced `xojs` function

https://github.com/xogroup/union/tree/master/packages/xojs/src/array/flatten

You can see we compared our choice to at least one other implementation (see the benchmark file)

GeorgeTaveras1231: With all of this said, this can be deferred. We are introducing `xojs/lib/callbacks/noop` in #186

hsbacot: @GeorgeTaveras1231 I don't think we should duplicate `lodash` functionality within `xojs`. It's redundant and the point is moot since `lodash` support selective imports as well. I think we should reconsider the inclusion of outsourced functions from `lodash` if for no other reason than the docs.

I agree with using it for other utility functions.

hsbacot:
@GeorgeTaveras1231 I can create a follow up for that. Where would you suggest it go? It's currently accessible by clicking on core components and finding it on the list.

GeorgeTaveras1231:
@hsbacot that is the 404 page which has become everyone's favorite way to navigate the docs lol. It should be accessible via the "Form Elements" page.

Issue #232, opened 2017-07-28
Introduce ToolTip


jmsardina:
We need to create a tooltip that can be either dismissible or non-dismissible.  The client should have control over the placement of the tooltip caret.

<img width="605" alt="screen shot 2017-07-28 at 3 29 17 pm" src="https://user-images.githubusercontent.com/10458903/28735136-910314d8-73a9-11e7-992b-8df65c07dd91.png">

<img width="870" alt="screen shot 2017-07-28 at 3 27 53 pm" src="https://user-images.githubusercontent.com/10458903/28735104-6ec755c8-73a9-11e7-9a3d-58bdcf3e1b8a.png">

PR #257, opened 2017-08-08
Introduce TkComponentTooltip

jmsardina:
# Open for documentation. Still need to add tests.

This commit introduces the tooltip component, which is simply a
container with an arrow.  This component is configurable to allow
the client to set the position, offset and size of the arrow.

Though the documentation includes examples of dismissible and progressive
tooltips, these versions leverage other Union components to provide
those additional experiences, as the component is not 'shipped' with
those features natively.

Github Issue:
[#232]

## Related Issue
[#232]

## Types of changes

- [x] Documentation

## Checklist:

## Reviewer Checklist


packages/@xo-union/tk-component-tooltip/src/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+  children: PropTypes.node,
+  /**
+  * Needs to include CSS unit
+  */
+  arrowSize: PropTypes.string
+};
+
+
+TkComponentTooltip.defaultProps = {
+  offset: 50,
```
peterkle: Similar to the progress bar component, perhaps this percentage should fractional: 0.0 ... 1

packages/@xo-union/tk-component-tooltip/src/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+TkComponentTooltip.propTypes = {
+  /**
+  * Color of the tooltip
+  */
+  backgroundColor: PropTypes.string,
+  /**
+  * Position of arrow point along the edge of tooltip (0-100)
+  */
+  offset: PropTypes.number,
```
peterkle: I would call this `arrowOffset` as not to confuse people into thinking the tooltip itself can be offsetted.

packages/@xo-union/tk-component-tooltip/src/index.jsx
jmsardina: I can change that, but I see that we're still multiplying the decimal by 100, which is the same as using the percentage.

docs/articles/pattern-library/content-patterns/tk-component-tooltip/demos/DismissibleTooltipDemo.jsx
```diff
@@ -0,0 +1,48 @@
...
+          style={{
+            width: '240px',
+            backgroundColor: colors['var-gray-1'],
+            color: colors['var-white'],
+            borderRadius: '6px'
+          }}
+        >
+          <span>DISMISSIBLE TOOLTIP</span>
+
+          <Icon
```
smanwaring: Use `{ IconButton }` from '@xo-union/tk-component-icons`

packages/@xo-union/tk-component-tooltip/src/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+    [position]: arrowPlacement,
+    [`border${arrowBasePosition}Color`]: backgroundColor,
+    [offsetPosition]: computedOffset,
+    borderWidth: arrowSize
+  };
+
+  return (
+    <div className={styles['tk-component-tooltip']} {...passThroughProps}>
+      <span className={styles.arrow} style={arrowStyles} />
+      <span className={styles.arrow} style={arrowStyles} />
```
smanwaring: This seems to be an extra arrow

packages/@xo-union/tk-component-tooltip/src/index.jsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from '@xo-union/tk-component-tooltip/lib/css';
+import colors from '@xo-union/tk-css-colors';
+
+export default function TkComponentTooltip({
```
smanwaring: Rename to ToolTip

packages/@xo-union/tk-component-tooltip/src/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+  backgroundColor,
+  arrowSize,
```
GeorgeTaveras1231: I'm curious about the need to override the arrow size. I expected this to be a static value.

packages/@xo-union/tk-component-tooltip/src/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+    minHeight: `calc(3 * ${arrowSize})`,
+    backgroundColor
+  };
+
+  const arrowStyles = {
+    [position]: arrowPlacement,
+    [`border${arrowBasePosition}Color`]: backgroundColor,
+    [offsetPosition]: computedOffset,
+    borderWidth: arrowSize
+  };
```
GeorgeTaveras1231: This should be tested

packages/@xo-union/tk-component-tooltip/src/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+      offsetPosition = 'left';
+  }
+
+  const boxStyles = {
+    // The mininum dimensions are set based on the arrow size to prevent
+    // the arrow from being larger than the tooltip box
+    minWidth: `calc(3 * ${arrowSize})`,
+    minHeight: `calc(3 * ${arrowSize})`,
+    backgroundColor
+  };
```
GeorgeTaveras1231: This should be tested

packages/@xo-union/tk-component-tooltip/src/index.jsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from '@xo-union/tk-component-tooltip/lib/css';
+import colors from '@xo-union/tk-css-colors';
+
+export default function TkComponentTooltip({
+  backgroundColor,
```
GeorgeTaveras1231: It is probably fine to accept this `backgroundColor`. Especially since your squad is still planning to experiment with this component. However, I'd like for there to be a number of standard colors (maybe 3) that are allowed for the tooltip. Similar to how the Button component is.

packages/@xo-union/tk-component-tooltip/src/index.jsx
jmsardina: The arrow size may be different sizes, but the consumer might not need/want to have to think about this.  I think we can have the arrow grow in relation to the size of the tooltip box so that the proportion is static but the size is not.
screen shot 2017-08-11 at 3 23 17 pm

packages/@xo-union/tk-component-tooltip/src/index.jsx
jmsardina: The extra arrow is supposed to help simulate a drop shadow, but I see I didn't commit that style.  Will update the branch.

packages/@xo-union/tk-component-tooltip/src/index.jsx
jmsardina: That makes sense.  I just have to run this by Design to see what colors should be allowed.

packages/@xo-union/tk-component-tooltip/src/index.jsx
hsbacot: Agree with @GeorgeTaveras1231 here. We should try to decide on one tooltip arrow size for all tooltips. Let's raise this with design

hsbacot:
Can the demo show the tooltip in action around a few elements to demonstrate positioning and how to control it from the parent since they will always be paired to another element?

Issue #250, opened 2017-08-03
Add Autoprefixer


smanwaring:
Context
===

- [ ] Visual bug?
- [ ] Runtime bug?
- [x] Build bug?

Issue
===

We need to add an Autoprefixer to Webpack to autoprefix our css.

marioharper:
This should also include going through and removing any hardcoded vendor prefixes.

PR #274, opened 2017-08-11
Feature/use html5 progress element

Samueljoli:
## Description

Change the underlying element used in `ProgressBar` from `div` to `progress`.

**NOTE:**
This PR is being opened prematurely. Mainly because I would like to open a discussion around my findings.

Some things to consider with using `<progress>`
1. Each browser provides it's own set of pseudo classes to style the status meter within the progress bar. So styling can only be applied via css. Inline styling will not work, therefore providing your own color is disabled, hence why that test has been skipped.
2. In our `scss` file we would have to account for each vendor prefix.
3. Support is ok across browsers and versions.

If there is any suggestions around being able to color the status meter I would greatly appreciate it 😄 

screen shot 2017-08-10 at 11 17 37 am

## Motivation and Context

Consistent semantic markup
## Types of changes

- [ ] Documentation

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/xogroup/union/blob/master/.github/CONTRIBUTING.md) document.
- [ ] I have updated the documentation as needed.

## Reviewer Checklist


Samueljoli:
@GeorgeTaveras1231 Hey, here's my findings. Maybe you have some suggestions around getting custom meter color functionality into this PR. Or maybe the effort isn't worth it.

GeorgeTaveras1231:
@Samueljoli we tried this in our end and it seems to work, but changing the `-webkit-progress-value`'s background color to `currentColor` allows consumers to overide the color via inline styles.

```scss
progress::-webkit-progress-value {
  background-color: currentColor;
}
```

Then the following works:

```html
<progress style="color: green;" value="0.5"></progress>
```

GeorgeTaveras1231:
Also, support for `currentColor` is pretty good: [Can I Use](http://caniuse.com/#search=currentColor)

packages/@xo-union/tk-component-progress-bar/src/components/ProgressBar/index.jsx
```diff
@@ -8,12 +8,21 @@ const ProgressBar = ({ progress, color }) => {
...
+      color="green"
+      className={styles.progress}
+      max="100"
+      value={progressWidth}
+    >
+
+      <div className={styles['progress-bar']} >
+        <div
+          className={styles['progress-fallback']}
+          style={{ width: `${progressWidth}%`, backgroundColor: color || null }}
```
hsbacot: use default props to default color to null

packages/@xo-union/tk-component-progress-bar/src/components/ProgressBar/index.jsx
```diff
@@ -8,12 +8,21 @@ const ProgressBar = ({ progress, color }) => {
...
 
   return (
-    <div className={styles['progress-bar']} >
-      <div
-        className={styles.progress}
-        style={{ width: `${progressWidth}%`, backgroundColor: color || null }}
-      />
-    </div>
+    <progress
+      color="green"
```
hsbacot: Should this be hardcoded to green?

packages/@xo-union/tk-component-progress-bar/src/components/ProgressBar/browser-spec.jsx
```diff
@@ -11,7 +11,7 @@ describe('<ProgressBar />', () => {
     return getProgressBar(props).find(progressSelector).node;
   };
 
-  it('should set the progress bar color', () => {
+  xit('should set the progress bar color', () => {
```
hsbacot: How come this test is pending?

Samueljoli:
@hsbacot @GeorgeTaveras1231 This PR is good to Review. I guess the only outstanding thing, is if your ok with accounting for all the vendor prefixes.

GeorgeTaveras1231:
@Samueljoli I am ok with that in the mean time, we opened #250 to simplify this in the future.

PR #276, opened 2017-08-14
Update RadioButton API to be similar to Checkbox.

marioharper:
## Description

Randomly generate RadioButton's id property. Remove `name` prop from `RadioGroup`. `name` prop will be covered when passing through props via `RadioButton`'s `{...props}` passthrough.

## Motivation and Context

To keep parity between related components and their API's. 

## Types of changes

- [ ] Documentation
- [x] Breaking change (fix or feature that would cause existing functionality to change. you had to modify existing tests.)
     - `RadioGroup` no longer accepts the `name` prop.

## Checklist:

## Reviewer Checklist


GeorgeTaveras1231:
@marioharper Thanks for the PR, But is it really necessary to randomly generate an ID? I understand the desire to simplify the API for the consumer, but how much work and how likely is it for the consumer to have ID conflicts? I think having an 'odd' name works for most scenarios (such as `__union-checkbox-{name}` ) and avoids the extra dependency.

marioharper:
The purpose was to keep uniqueness while not having a required dependency on the `name` prop.  Since `name` is only **necessary** when sending formdata. `_union-checkbox-{name}` would create that dependency on `name`. Now, adding the `name` prop will be the same as a vanilla `<input type="radio" name="my-name" />` just add it to each individual `RadioButton`

marioharper:
Also, `_union-checkbox-{name}` would create multiple id's with the same value, since each `RadioButton` shares the same `name`.

GeorgeTaveras1231:
@marioharper but I think that the dependency on the name prop is better than the dependency on `shortid`. Also, does that mean that if you want to send the formdata, you need to pass the name to each checkbox? That seems pretty tedious. I also think that imposing the default behavior of sending form data is reasonable since it provides a mechanism for fault tolerance (When server-rendering a form, if the js can't be loaded for some reason, you can still submit the form.)

marioharper:
@GeorgeTaveras1231 I agree there pros and cons to each approach. Another reason we went for this approach in `Checkbox` was because if you make `name` required on `checkbox` but the consumer just supplies `name` on `CheckboxGroup` and allows `CheckboxGroup` to set the name of each child checkbox component, you will always get a `PropTypes` warning for not setting the required `name` property on each child `Checkbox`.  `PropTypes` isn't smart enough to know that `name` is getting set during cloning of each child.

marioharper:
@smanwaring @hsbacot what do you guys think?

GeorgeTaveras1231:
@marioharper one thing I failed to point out yesterday is that adding the generation of shortid in the render function makes the `RadioButton` component non-idempotent, which will lead to unnecessary re-renders (as a new ID will be generated everytime the render function is called).

If we MUST use a dynamic id, we should make `RadioButton` a class component and create the id in the constructor.

smanwaring:
@marioharper I spoke with George this AM and he commented above with what we thought would make the most sense given our concerns.

GeorgeTaveras1231:
@marioharper I'm also curios about the console warnings you mentioned. I feel that it shouldn't be a problem as long as the `name` prop is not marked as `required`

hsbacot:
I agree with the introduction of `shortid` to prevent duplicate id's.

hsbacot:
This shouldn't create any additional renders because it will only be created while rendering the `RadioButton`

marioharper:
Yeah I don't think this will cause re-renders. Like @hsbacot pointed out, it's called in the render method so it won't cause a re-render. There's no state change that would trigger a re-render.

hsbacot:
I think this is good to go 👍

I'm approving but I'd wait to merge until others are in line with this.

GeorgeTaveras1231:
@marioharper @hsbacot The state change doesn't need to happen in the component itself. It can happen in any of it's parents. 

Look at what happens when I change the state of the `FormDemo` component

aug-15-2017 12-27-23

marioharper:
@GeorgeTaveras1231 doesn't that re-renders because the `checked` property changed on the `checkbox`, which caused a re-render. Having the randomly generated id doesn't affect this. It just visually shows you that that entire `checkbox` re-rendered due to a prop change.

GeorgeTaveras1231:
@marioharper That is not true. Yes, the `checked` prop of the `Checkbox` is changing, which should cause a partial re-render of the `Checkbox` component. However, nothing has changed in the `RadioButton` component yet it is re-rendering. If you hard code the `id`, the virtual DOM would detect that the ui for the `RadioButton` has not changed since the last render, which should lead to no re-rendering of the `RadioButton`.

To further demonstrate, this is how it looks when changing `RadioButton` to a class and using an instance property for the `id`.

aug-15-2017 14-30-13

marioharper:
ohhhhh, that's my mistake, I was looking at the completely wrong thing. I didn't realize your screencap was showing the `RadioButton` changing when changing a `Checkbox`'s state... I thought it was just showing the checkbox

GeorgeTaveras1231:
@marioharper @hsbacot This is not so bad in our demo page but image this component being a child of a controlled form with `input[type="text"]` inputs. Literally, as the user is typing, the `RadioButton` will just go bananas 🍌

GeorgeTaveras1231:
BTW @marioharper, I updated the branch since there is a bug that was fixed in master that could wipe out the colors package lol

marioharper:
@GeorgeTaveras1231 Thanks for teaching me something new about how react re-renders! I'm cool with making it a class. This will have to be done for `Checkbox` as well, since this is currently how checkbox is implemented and it has the same re-render issue. I'll push up a change once I get to a good stopping point on location typeahead.

GeorgeTaveras1231:
@marioharper No worries, thanks for being receptive to feedback. 

These issues are easier to catch when you create a mental model of render functions always being **idempotent** or thinking of them always having **referential integrity**. Both things meaning that with a given `this`, `props` and `state`, `#render` should **always** return the same element tree. This makes any **non-deterministic** code bad for `#render` functions, things like `Math.random()`, `new Date()`, etc.

Also, I ran into this same issue when building out the `Field` component. One of the original implementations used a randomly generated ID as well. 😄

marioharper:
@GeorgeTaveras1231 just pushed up the changes, should be good after the ci runs.

packages/@xo-union/tk-component-fields/src/components/RadioButton/index.jsx
```diff
@@ -1,30 +1,37 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { PropTypes } from 'prop-types';
 import RadioButtonCss from '@xo-union/tk-component-fields/lib/css';
+import shortid from 'shortid';
 
-const RadioButton = ({ children, value, name, checked, onChange }) => {
-  const id = `${name}-${value}`;
+class RadioButton extends Component {
+  state = {
```
GeorgeTaveras1231: I feel that using `state` implies that it will change. Why not just use an instance property?

```jsx
class RadioButton extends Component {
  id = shortid.generate()
}
```

hsbacot:
I'm not convinced that render isn't being called in your example, rather it isn't as obvious because the key isn't randomly generated on each render. I think the render is coming from `React.Children.map`

marioharper:
It is being called, but nothing changes so the DOM won't actually update.

GeorgeTaveras1231:
@marioharper I approved but still opened #296 to propose replacing the `shortid` dependency for other reliable methods to generate unique ids.

marioharper:
@GeorgeTaveras1231 going to merge this through and will address the shortid dep separately

Issue #279, opened 2017-08-14
Document Column component


GeorgeTaveras1231:
This component currently lives in `@xo-union/tk-component-grid`

PR #286, opened 2017-08-15
Document grid components

GeorgeTaveras1231:
Closes: #279


docs/articles/pattern-library/core-components/grid/demos/StackedToHorizontal.jsx
```diff
@@ -16,48 +17,48 @@ function createSmallColumn() {
...
+      </Column>
+    </Row>
   );
 }
 
 
 export default function StackedToHorizontal() {
   return (
-    <div className={styles['my-container']}>
+    <Container>
```
smanwaring: None of these examples are actually becoming stacked in the demo. I can't immediately see why that would be happening. When I inspected, it looks like the columns are also getting a 'col' class, which if applied, would explain why they aren't stacking.

docs/articles/pattern-library/core-components/grid/README.md
```diff
@@ -23,9 +31,11 @@ This is a CSS Module. Check the [getting started with CSS Modules](/pattern-libr
 ## Install
 
 <InstallSnippet packageJson={packageJson} />
+<InstallSnippet packageJson={componentsPackageJson} />
 
 ## Introduction / How it works
-The Union grid system is a CSS modules wrapper to Twitter Bootstrap (v4); however, the modules have been configured in a custom way. Many of the predefined grid classes that come from Twitter Bootstrap are available.
+
```
smanwaring: The top of the documentation page says that it is a CSS module, which is a little confusing.

PR #296, opened 2017-08-15
Remove unecessary dependency "shortid"

GeorgeTaveras1231:
I want to show a path for generating unique IDs without requiring a dependency.

Additionally, I updated the `Field` `Dropdown` and `Textarea` component to use the same method for a unique id.


packages/@xo-union/tk-component-fields/src/utilities/index.js
```diff
@@ -1,9 +1,12 @@
...
   return string.split(/[\W\s]/g).map(word =>
     word.replace(/^(\w)/, w => w.toUpperCase())
   ).join(' ');
 }
 
-export function fieldId(name) {
-  return `__ff-${name}__`;
+let idCounter = 0;
+export function fieldId() {
+  return `__UNION_FIELD__${__webpack_hash__}_${idCounter++}__`;
```
GeorgeTaveras1231: @marioharper This being the magic sauce that generates unique IDs. It uses the webpack hash which will be unique per package version. And using an incrementing counter.

GeorgeTaveras1231:
History needs to be cleaned since #276 was merged, which was the original base for the PR

marioharper:
@GeorgeTaveras1231 is the webpack hash necessary?

GeorgeTaveras1231:
@marioharper Yes to account for the edge case where two versions of this same package make it into the same bundle. Something like

```
my-app
|___ @xo-union/tk-component-membership-modal@1.0.0
|     |___@xo-union/tk-component-fields@2.0.0
|___@xo-union/tk-component-fields@1.0.0
```

GeorgeTaveras1231:
@marioharper otherwise, a counter would do. But because the counter is instantiated per instance of the module, it would be problematic with multiple instances

PR #311, opened 2017-08-17
Document membership libraries

GeorgeTaveras1231:


hsbacot:
Could we mention earlier in that docs that there are two types of configuration supported and this is an enhancement? Perhaps we include something that it isn't necessary to migrate to the webpack configuration as well.

GeorgeTaveras1231:
@hsbacot thanks for that feedback. I went ahead and pointed out that there are two approaches higher in the documentation page (right below the **Usage**). However, I would rather not say that using the environment variables isn't required because there isn't a strategy to globally configure all instances of the client that does not use the environment variables. I would rather defer this feature until a later time. At least for now, there is a path to configure all clients.

docs/articles/pattern-library/utilities/membership-sdk/README.md
```diff
@@ -0,0 +1,137 @@
...
+    host: 'my-host',
+    apiKey: 'my-api-key'
+  }
+});
+```
+
+
+### Methods
+
+#### `#signUp` => Promise
```
marioharper: Could we make this clearer that the Promise resolves with an object containing `member` and `session`? I know its documented below, but it seems like that should be documented here.

docs/articles/pattern-library/utilities/membership-store/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+<InstallSnippet packageJson={packageJson} />
+
+## Pre requisites
+
+Most users of this library will need configure their `CoreProvider`. Follow the instructions [here](/pattern-library/getting-started/global-dependencies#state).
+
+## Usage
+
+To configure the MembershipService, follow the instructions [here](/pattern-library/utilities/membership-store)
```
marioharper: should this point to `/pattern-library/utilities/membership-sdk`?

